PR #6, opened 2016-07-31
leixperiment sync

leilei-:
This merges most of the (stable-ish) crazy stuff I've developed over the past two years. Diffed by hand, though I may have carried over a few old strcmps here and there by mistake.

It also has the lfx particle trap changes (with all the old cgame particle stuff ripped out), so it would be advised to set cg_leiEnhancement to 0 before starting a game to avoid invalid trap errors.

I've also fixed qvm compiling for possession.  I have played 3 rounds of possession on the new OA3 data now :)

It probably needs fixes to not make travis upset, i've only had tried compiling with the lcc/q3asm stuff without issue on short notice..


sago007:
I had to add a 
`#undef NUM_CROSSHAIRS` to get the qvm files to compile.

The non-qvm build fails. 

I applied these changes to get it to compile:
[sago_patch_2016-07-31.patch.zip](https://github.com/OpenArena/gamecode/files/392348/sago_patch_2016-07-31.patch.zip)

It it properly wrong (I don't understand what the eyepos-array holds) but it makes it compile.

leilei-:
eyepos is the location of the eye relative to the head model origin, where it's given in eyes.cfg.  It could be refactored to renderer code later on for mod compatibility

sago007:
I looked at it, I suggest merging it too. There are a few indentation irregularities but nothing important. It is a general problem anyway.

sago007:
Do you plan on merging more or can I start looking through the code?

There are a lot of unused variables and I might have an easier time cleaning them up as I appear to have a slightly more aggressive compiler.

leilei-:
I'm shifting back to doing art, so i'm done for now.  What's not merged from the leixperimental at this point is some monster stuff and precaching of now-unused particle shaders

Issue #7, opened 2016-10-31
cg_bobup is cheat protected


ldrone:
Is the any reason cg_bobup should be a cheat variable and not changeable by client?

leilei-:
Because you can clip your view through the floor and cheat with it.

ldrone:
I've also noticed there is a sway in view when hit by a rocket on the side.
Would it be simpler to implement a bool variable to disable/enable all bob/roll?

leilei-:
i'd propose these:

cg_bob = a bool that can allow or skip the bobup/pitch/roll bit in cg_view.c
cg_kick = a float to adjust intensity of the view kicking, in the same function

The-Gig:
Excuse me, is this "bob" thing the very small camera change that happens when you start or stop moving??
I tried setting cg_bobup to much higher values (default value is 0.005), and it looked like the camera height when start ot stop moving changed a bit.
But also, I had the impression that, while walking with such higher values, the world was "flickering" a little at times. :-/
What's the goal of changing these settings?

And about view kicking (when we are hit by a rocket?), I have noticed this (not merged yet) commit: https://github.com/OpenArena/gamecode/pull/9/commits/08d47463aa81a128e724dc4d241c779bbe18b89f
Isn't it altering the current behavior of what's that related, by adding a multiplier which is set to 0.002 by default? Why a such change? For having the original behavior, the multiplier has to be set to 1.000?
I'm not used to C, **so please have patience with me when I will make stupid questions...**

ldrone:
The goal is if setting cg_bob to 0 it would disable all the bobbing (the camera movement).

Yes, cg_kickscale should be default to 1. Forgot to change it. Setting it to 0 would disable you screen tilting (rotating) when hit by a rocket. Setting it to other values would adjust the effect, although I don't know why would anyone want to do this.

PR #9, opened 2016-11-02
View changes

ldrone:
Set the cg_kickScale to 1 by default.


sago007:
As far as I can see the code is good and implements the solution @leilei- suggested in #7 perfectly.

The-Gig:
I tried the feature with the current nightly build.
It's okay, but just a little thing: If I understood correctly, it supports decimal parts to fine-tune the feature (e.g. setting it to 1.0, 1.5 or 1.9 changes the result), right?
But the default value the game mentions is "1". Is it possible to modify it to show a default of "1.0", to allow users to know the variable supports decimals and not just integer?

PR #10, opened 2016-11-07
Integratied UI3 into the gamecode repository to allow nightly builds.

sago007:
For a long time I wanted to try the UI3. I thought it was a bit hard so I think including it in the nightly builds would be a good idea. However this means that the UI3 files need to be there.

There are two options as I see it:
Include the UI3 directly in the gamecode and remove the old UI3 repository. 
Alternatively add the UI3 repository as a sub repository.

I prefer the first approach which is also what is contained in this repository. Not only because a gitsub repository is not yet that user friendly but also for the following reasons:

I believe it makes sense to include it directly in the gamecode since you will often need to change the menu along with the gamecode.

Of course UI3 also contains some binary files. They should properly be removed the day that they are no longer needed. But I don't think they are that big compared to how much easier it will be to checkout and compile a working UI3 copy.

Any opinions?


The-Gig:
My humble opinion about allowing us to test UI3 is surely positive.
- I'm curious to try it.
- People will be able to provide some feedback about it.
- When you will implement something "new", you will not need anymore to also update the old ui (I'm a bit fond to the old one, but Fromhell said with the new one it's much easier for you to modify menus). A recent exampe is the new "possession" gametype... the time it took to you to make the required changes to the old ui may have been spent elsewhere.
- It's a step towards OA3 which, sooner or later, will be required.

About the best way to manage that (e.g. a separate pk3 with thenecessary binary stuff -e.g. icons etc- to be updated less frequently, or other solutions), **I don't know...**, sorry.

PS: Of course, I have absolutely no idea about how much "complete" the new UI is now. It is possible it is still lacking important stuff...

leilei-:
Yeah i'm thinking the scripts only too, for the sake of a code repository staying a code repository

The binary stuff in the UI3 is there for an all-in-one test and is considered OA3 base pak0.pk3 material. and i'm also afraid of the fragmenting from testing (i.e. servers pushing outdated UI3 test paks later on)

The-Gig:
I understand your fear about people potentially misusing the OAX test packages (for example, that's the reason why, when we were making OACMP Vol1, we did not switch to "final" map names until a late stage of development), but on the other hand, we cannot delay oa3 ui development forever....

A thing I can think about: I suppose also in the new UI there will be the game version in the main screen, right? In OAX, that's just a small "OAX" test in a corner. What about adding also a big red warning there, something like "THIS IS OAX, DEVELOPMENT TEST VERSION OF OPENARENA. DO NOT USE THIS IN ANY REAL SERVER". People re-using OAX OA3 ui would also carry that bulky test...

PS: I don't really know about the sdl input controversy (I wasn't yet around OA back then)... But I think I hear someone mentioning it before... Was it something which changed the controls "felling"?

sago007:
I don't think outdated test packs are going to be a problem. At least not bigger than the current risk of problematic pk3s. I did have the same fear about oax releases. That is the reason I always ships them in a uniquely named subfolder.

While I do not like binary files in the code repository. But I do believe that all files necessary to build and run should be present. 

I can add the necessary script to the UI3 repository instead to generate the binary pk3. But I don't know if they are static enough for this approach. They seem tightly connected to me. If they need to be developed together they might need to stay together. At least until OA3 (or another patch) is out.

sago007:
I'll attack the problem differently

The-Gig:
Excuse me Sago, what do you mean?

The-Gig:
I guess this is the answer, for the moment, right?
http://openarena.ws/board/index.php?topic=5296.0

Issue #8, opened 2016-10-31
Disable obituary output to console


ldrone:
Playing a game of insta ctf mod, frag messages appear in small intervals, clogging the console and making the chat history unreadable. The same probably applies to deathmatch mode.

I am thinking of implementing a variable cg_fragmsgoutput [0|1], zero disabling the output of obituary messages to console. The frag message would still appear on the hud (upper left corner).

The-Gig:
Just a question: are those infos used by statistics analysis tools?
http://openarena.wikia.com/wiki/Servers#Server_statistics

ldrone:
I'm referring to messages of the type: "player_x almost dodged player_y's rocket", or other info such as
player_x got the flag. These are all shown on the top left screen and written in console.

My idea is to have an option not write them into console. If those are used for statistics, they can be enabled on the server side, but I'd have it set default to 0 for clients.

The-Gig:
"Default to 0" does not seem a good idea to me. For me it's hard to read messages in top left corner,  finding everything in console output is useful IMHO.

Another question... when there is a single cvar which has effect on both client and server aspects... does its prefix begin with com_ , usually? Just asking...

ldrone:
Anything not having sv_ prefix  is not server related. So everything else, in theory, should be client side. I'm not sure I completely understand your question. How does a cvar affect both client and server? Give me an example.

For a list of prefixes, from Quake Live [http://www.regurge.at/ql]():
BOT_: bot settings
CG_: client game settings
CL_: client-side settings
CM_: collision map settings
COM_: Common settings
FS_: game files settings
G_: server-side game settings
GT_: connection settings
IN_: general input device settings
JOY_: joystick input settings
M_: mouse input settings
NET_: network settings
PMOVE_: player movement server settings
R_: video rendering settings
S_: sound system settings
SV_: server-side settings
SYS_: system configuration settings
UI_: user interface settings
WEB_: website settings

The-Gig:
I was thinking about "common settings" (com_) due to the console sounds something where both client and server parts of the game write output.
**I don't know** exactly if the "feature" you are thinking about should be considered a client one or a common one... So I asked. Would the command also affect a dedicated server, if typed there?

PS: Thank you for the list, that's interesting!

leilei-:
games.log logs the kills iirc.  and personally i'm hoping for icongraphic obituaries ala Half-Life for the UI3 system at least but i've never messed with cgame string printing all that much.

[...]

PR #11, opened 2016-11-20
Frag message output with icons

ghost:


The-Gig:
Just adding a link to the previous discussion on the topic:
https://github.com/OpenArena/gamecode/issues/8

The-Gig:
Excuse me Idrone, you did not answer to a question I did here:
https://github.com/OpenArena/gamecode/issues/8#issuecomment-261703510

PS: You already know, I don't see any reason for hiding the attacker name, in case of indirect kill.

PPS: just asking for curiosity... why does the "case MOD_BFG:" section is completely empty? Because so it automatically takes the same settings as "case MOD_BFG_SPLASH" which is right under it? You know I'm not good with C...

The-Gig:
Another thing: when someone is killed by a teammate, should't the iconographic version mention this, too? Like "x <skull icon> y (TEAMMATE)"...

ldrone:
> PPS: just asking for curiosity... why does the "case MOD_BFG:" section is completely empty? Because so it automatically takes the same settings as "case MOD_BFG_SPLASH" which is right under it? You know I'm not good with C...

It "falls through" to the next case since there is no `break;` statement, which would exit the switch and resume after it. So the compiler continues to read the next lines of  `case MOD_BFG_SPLASH` until the next `break;` is encountered.

The-Gig:
About the bfg question... so it's like I guessed. Thank you!

sago007:
Ok, I have checked it out and also tryed playing it. Looks good. A bit hard to read on my screen but I could not read the console either, so it does not matter.

Regarding the code:
Your code is indented with spaces while most of the code is indented with tabs. I must admit that I have done the same at some point. Github just makes it much more obvious. I am not taking side in the spaces vs tab argumentation but it should be the same in all of the code. 

Also something like this:
```
 Q_strncpyz(lastFrag->targetName, targetName, 32);
```
The "32" should either be replaced by a "#define MAX_FRAG_NAME_LENGTH" or a "sizeof".

```
static int fadeTime = 7000;
```
should properly be
```
const int fadeTime = 7000;
```

I would greatly prefer curly brackets here:
```
if (cgs.fragMsg[i].teamFrag)
            CG_DrawStringExt(x + x_offset, y + (i * TINYCHAR_HEIGHT),
                  "(^1TEAMMATE^7)", hcolor, qfalse, qfalse,
                  TINYCHAR_WIDTH, TINYCHAR_HEIGHT, 0);
```
While it is only 1 logical line it is 3 physical lines and I would mark it. You also used curly brackets other places where the if content was more than 1 physical line.

Also
```
            if (ent->generic1) lastFrag->teamFrag = qtrue;
            else lastFrag->teamFrag = qfalse;
```
could be
```
            if (ent->generic1) {
                lastFrag->teamFrag = qtrue;
            }
            else {
                lastFrag->teamFrag = qfalse;
            }
```
or maybe just
```
lastFrag->teamFrag = ent->generic1;
```

Aside from that the code looked solid. 

Perhaps the commits should be squashed before merging. While it is easy enough to read all the commits as one in this pull request it might be harder in the commit history.

ldrone:
I changed the cg_kickScale to show 1.0 as Gig suggested in #9. Should I leave that here since it's a minor change or else how would I go to make that a separate commit?

sago007:
Doesn't matter. Just let it be in this.

ldrone:
Replaced spaces with tabs. Currently in my editor, it's hard to see how the tabs between words will look on github.

sago007:
Looks good to me!

PR #13, opened 2016-12-27
Frag msg v2

ldrone:


ldrone:
This code for testing empty string as sago007 suggested, isn't working for some reason:
```
if (fi.attackerName != '\0')
```

sago007:
> if (fi.attackerName != '\0')

I was thinking more like this:
`if (fi.attackerName[0] != '\0')`
We test if the first char is the end.
This should properly have a function or a macro.

ldrone:
My mistake.

sago007:
Looks good to me

The-Gig:
For completeness, considering this commit reorganized the options:

**cg_obituaryoutput**:
- 0: no obituary output
- 1: console (text) only ("classic" behavior)
- 2: hud (text) only
- 3: hud (icons) only (new default)
- 4: hud (icons) + console (text)

PR #14, opened 2016-12-28
Realigned the images for frag message display.

ldrone:


sago007:
I like that it removes more lines than it adds.

PR #17, opened 2017-06-24
Be generically portable to new CPUs on all GNU platforms

smcv:
This pull request collects several portability changes originating in Debian that were merged into ioquake3 between 2011 and 2015.

* Treat GNU/kFreeBSD and GNU/Hurd as essentially equivalent to GNU/Linux, allowing compilation on those platforms. OpenArena already had part of the kFreeBSD change due to syncing the Makefile from ioquake3, but not enough to make the build work there.
    * ioquake3 pull request: https://github.com/ioquake/ioq3/pull/4 (that's only the Hurd part, the kFreeBSD parts were merged while ioquake3 was still in svn)
* Treat all CPUs on GNU platforms as essentially equivalent, avoiding needing to add specific support for each new CPU. In most cases the `ARCH_STRING` is just the output of `uname -m`. The only special cases needed are for 32-bit x86 and ARM, where `uname -m` can take multiple values (such as `i586`, `i686`, `armv5tel`, `armv7l`) which we need to normalize to `x86` and `arm` to be compatible across CPUs from the same family. This patch is enough to avoid needing a change like https://github.com/ioquake/ioq3/pull/128 for each new architecture, and in particular provides aarch64 (64-bit ARM) support.
    * ioquake3 pull request: https://github.com/ioquake/ioq3/pull/129

The `ARCH_STRING` is not actually used anywhere in the gamecode, except for raising a `#error` if it isn't defined. I could have deleted it completely, but I thought it was probably better to keep `q_platform.h` consistent with the one in ioquake3.

Please see the commit messages of the individual commits for more information.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -6,7 +6,7 @@
 
 COMPILE_PLATFORM=$(shell uname|sed -e s/_.*//|tr '[:upper:]' '[:lower:]'|sed -e 's/\//_/g')
 
-COMPILE_ARCH=$(shell uname -m | sed -e s/i.86/x86/)
+COMPILE_ARCH=$(shell uname -m | sed -e s/i.86/x86/ | sed -e 's/^arm.*/arm/')
 
 ifeq ($(COMPILE_PLATFORM),sunos)
   # Solaris uname and GNU uname differ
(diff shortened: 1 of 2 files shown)


sago007:
Looks good to me

PR #18, opened 2017-06-24
Pick up date from SOURCE_DATE_EPOCH, for reproducible builds

smcv:
The goal of reproducible builds is that a rebuild of the same source code with the same compiler, libraries, etc. should result in the same binaries. SOURCE_DATE_EPOCH provides a standard way for build systems to fill in the date of the latest source change, typically from a git commit or from metadata like the debian/changelog in Debian packages.
    
This does not change anything if SOURCE_DATE_EPOCH is not defined; the intention is that a larger build system like a Debian package will define it.
    
Please see https://reproducible-builds.org/ for more information about reproducible builds.
    
Most of this pull request was sent to ioquake3 as commit https://github.com/ioquake/ioq3/commit/9c76b546e33559c5679f6c066ff976bc7f923cbb (which also touched engine files not present here). This version also defines `PRODUCT_DATE` when invoking `q3lcc`, making `qagame.qvm` reproducible if a suitable `SOURCE_DATE_EPOCH` is provided.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1087,6 +1087,11 @@ else
   STRIP_FLAG = -s
 endif
 
+# https://reproducible-builds.org/specs/source-date-epoch/
+ifdef SOURCE_DATE_EPOCH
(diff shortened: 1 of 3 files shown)


sago007:
Looks good to me

PR #19, opened 2017-07-25
Added cg_zoomSensitivity for correct sens scaling

ryvnf:
The zoom sensitivity scaling formula in OpenArena is `vertical_zoomFov / 75.0`.

This formula is flawed in 2 ways:

1. It assumes that the vertical unzoomed fov is 75. This is eqvivalent to
`cg_fov` being set to ~91 on 4:3 or ~107 on 16:9.

2. It scales the zoom sensitivity linearly using `vertical_zoomfov /
vertical_unzoomedfov`. This is mathematically incorrect which is noticable
through testing.

New cvar called `cg_zoomSensitivity` has been added to solve these problems.
When it is set to 0 (default) it has no effect and the legacy zoom sensitivity
formula is used.

When the cvar is set to anything but 0 the mathemtically correct formula
`tan(zoomFov / 2) / tan(unzoomedFov / 2)` is used. The result of this formula
gets multiplied by `cg_zoomSensitivity`.

When the cvar is set to 0.5 for example it will multiply the result of the
formula with 0.5 which halves the zoom sensitivity. This allows players to
customize their zoom sensitivity according to their preference.

### DIFF
--- a/code/cgame/cg_local.h
+++ b/code/cgame/cg_local.h
@@ -1334,6 +1334,7 @@ extern	vmCvar_t		cg_ignore;
 extern	vmCvar_t		cg_simpleItems;
 extern	vmCvar_t		cg_fov;
 extern	vmCvar_t		cg_zoomFov;
+extern	vmCvar_t		cg_zoomSensitivity;
 extern	vmCvar_t		cg_thirdPersonRange;
 extern	vmCvar_t		cg_thirdPersonAngle;
 extern	vmCvar_t		cg_thirdPerson;
--- a/code/cgame/cg_main.c
+++ b/code/cgame/cg_main.c
@@ -147,6 +147,7 @@ vmCvar_t cg_ignore;
 vmCvar_t cg_simpleItems;
 vmCvar_t cg_fov;
 vmCvar_t cg_zoomFov;
+vmCvar_t cg_zoomSensitivity;
 vmCvar_t cg_thirdPerson;
 vmCvar_t cg_thirdPersonRange;
 vmCvar_t cg_thirdPersonAngle;
@@ -308,7 +309,8 @@ static cvarTable_t cvarTable[] = {// bk001129
 	{ &cg_ignore, "cg_ignore", "0", 0}, // used for debugging
 	{ &cg_autoswitch, "cg_autoswitch", "1", CVAR_ARCHIVE},
 	{ &cg_drawGun, "cg_drawGun", "1", CVAR_ARCHIVE},
-	{ &cg_zoomFov, "cg_zoomfov", "22.5", CVAR_ARCHIVE},
+	{ &cg_zoomFov, "cg_zoomFov", "22.5", CVAR_ARCHIVE},
+	{ &cg_zoomSensitivity, "cg_zoomSensitivity", "0", CVAR_ARCHIVE},
 	{ &cg_fov, "cg_fov", "90", CVAR_ARCHIVE},
 	{ &cg_viewsize, "cg_viewsize", "100", CVAR_ARCHIVE},
 	{ &cg_viewnudge, "cg_viewnudge", "0", CVAR_ARCHIVE},
--- a/code/cgame/cg_view.c
+++ b/code/cgame/cg_view.c
@@ -24,6 +24,8 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 // for a 3D rendering
 #include "cg_local.h"
 
+#define degrees(v) (v * 180.0 / M_PI)
+#define radians(v) (v * M_PI / 180.0)
 
 /*
 =============================================================================
@@ -638,9 +640,9 @@ static int CG_CalcFov( void ) {
 		fov_x = fov_x * 0.93 * (cg.xyspeed * (0.0006) + 1);
 	}
 
-	x = cg.refdef.width / tan( fov_x / 360 * M_PI );
+	x = cg.refdef.width / tan( radians(fov_x) / 2.0 );
 	fov_y = atan2( cg.refdef.height, x );
-	fov_y = fov_y * 360 / M_PI;
+	fov_y = degrees(fov_y) * 2;
 
 	// warp if underwater
 	contents = CG_PointContents( cg.refdef.vieworg, -1 );
@@ -664,7 +666,14 @@ static int CG_CalcFov( void ) {
 		cg.zoomSensitivity = 1;
 	} 
 	else {
-		cg.zoomSensitivity = cg.refdef.fov_y / 75.0;
+		float zoomSensitivity = cg_zoomSensitivity.value;
+		if (zoomSensitivity == 0.0) {
+			// use legacy zoom sensitivity formula
+			cg.zoomSensitivity = cg.refdef.fov_y / 75.0;
+		} else {
+			// use corrected zoom sensitivity formula
+			cg.zoomSensitivity = zoomSensitivity * tan(radians(cg.refdef.fov_x) / 2.0) / tan(radians(cg_fov.value) / 2.0);
+		}
 	}
 
 	return inwater;


ryvnf:
Check [this repository](https://github.com/OAGuild/zoomsens) to see results from testing that shows how much better the corrected zoom sensitivity formula is compared to linear zoom sensitivity scaling.

The-Gig:
Three questions:
1) Which value would you suggest players to use? Or di you have various suggested values depending from their cg_fov or screen ratio?
2) Wasn't there also a cg_zoomfov cvar? I'm not sure and I can't check now.
3) This change is in gamecode, right? How does it relate with the change in widescreen adaption (from vert- to hor+) that will differentiate OA 0.8.x engine and oa3 engine binaries?

ryvnf:
@The-Gig 

1. Which value would you suggest players to use? Or do you have various suggested values depending from their cg_fov or screen ratio?

I would suggest players to use the value `1.0`. This makes so that the zoom sensitivity while zoomed in feels exactly the same as while zoomed out (see [this](https://github.com/OAGuild/zoomsens). The value does not change depending on `cg_fov` or aspect ratio.

2. Wasn't there also a cg_zoomfov cvar? I'm not sure and I can't check now.

Yes, and it is still there. But I changed to capitalization from `cg_zoomfov` to `cg_zoomFov` because I thought it looked more consistent with the capitalization of `cg_zoomSensitivity`. The cvar-names are case-insensitive so this only affects the default capitalization of the variable when printed in console.

3. This change is in gamecode, right? How does it relate with the change in widescreen adaption (from vert- to hor+) that will differentiate OA 0.8.x engine and oa3 engine binaries?

Yes, this change is in gamecode. Widescreen adaption should not cause any problems with this change.

The-Gig:
So, this is about having the same "mouse sensitivity" ("mouse feeling") when you are zooming and not (side note: you did not mention "mouse" in this page, so to understand what it was about I had to think a little and see your page here: https://github.com/OAGuild/zoomsens).
But I haven't yet understood if the change affects the mouse sensitivity only in the _short moment_ when the zoom is _moving in or out_, or for the whole time you have the zoom active.

Also, in that page you linked with more infos, you say about the need to switch from just binding "+zoom" to binding a "toggle" between two values (which ones, specifically?). Could you please explain more that part? Do you mean that your cg_zoomSensitivity cvar would have absolutely no effect if the user is using the standard "+zoom" method or what? And in that case, for which reason it's not technically applicable also to that?

Sorry for annoying you, but I'm slow at understanding, at times...

The-Gig:
You said
> [The classic sensitivity scaling forumula] assumes that the vertical unzoomed fov is 75. This is eqvivalent to
> cg_fov being set to ~91 on 4:3 or ~107 on 16:9.
Does the part "~107 on 16:9" refers to OA 0.x engine (vert-) or OA3 engine (hor+)? From my little understanding about how that engine change works, I guess that refers to the "classic" engine, because with the new one the vertical fov will be the same (if cg_fov is the same) for both 4:3 and 16:9 (while on the other hand, horizontal fov will be actually larger on 16:9).
So, with the new engine, that sentence could be "This is eqvivalent to cg_fov being set to ~91 (on both 4:3 or 16:9).". Am I right or am I missing something?

ryvnf:
> Does the part "~107 on 16:9" refers to OA 0.x engine (vert-) or OA3 engine (hor+)? From my little understanding about how that engine change works, I guess that refers to the "classic" engine, because with the new one the vertical fov will be the same (if cg_fov is the same) for both 4:3 and 16:9 (while on the other hand, horizontal fov will be actually larger on 16:9). Right?
So, with the new engine, that sentence would be "This is equivalent to cg_fov being set to ~91 (on both 4:3 or 16:9).". Am I right or am I missing something?

cg_fov ~107 on 16:9 refers to OA 0.x (vert-)

You are right that there would be no difference of `cg_fov` between 4:3 and 16:9 and that `cg_fov` would be ~91 independent of aspect ratio.

You are not missing something as I can see.

Just to be clear: The number talked about here is the number where the sensitivity scaling of the "classic" formula is optimal. If you deviate from this number the sensitivity / "feel" will become a quite weird because the formula assumes that the vertical unzoomed fov is 75.

ryvnf:
> So, this is about having the same "mouse sensitivity" ("mouse feeling") when
> you are zooming and not (side note: you did not mention "mouse" in this page,
> so to understand what it was about I had to think a little and see your page
> here: https://github.com/OAGuild/zoomsens). This looks interesting: I have to
> admit I sometimes felt mouse sensitivity a bit different (mouse moving a
> little slower) while zooming... I thought that may have been on purpose to
> give you more accurate aiming... but maybe it's not on purpose. Now that I
> check, maybe the difference is more noticeable if I have cg_fov 120 instead of
> 90... however mine are just "sensations", I cannot tell "scientifically"...
>
> In that page you linked with more infos, you say about the need to switch from
> just binding "+zoom" to binding a "toggle" between two values (which ones,
> specifically?). Could you please explain more that part? Do you mean that your
> cg_zoomSensitivity cvar would have absolutely no effect if the user is using
> the standard "+zoom" method or what? And in that case, for which reason it's
> not technically applicable also to that?
>
> Sorry for annoying you, but I'm slow at understanding, at times...

Yes, I also felt that the "feel" while zoomed in isn't the same as when zoomed out. This is what lead me to research the problem a bit more, leading up to this fix.

The effect you are talking about increases the more you deviate from 75 vertical fov. Do `set cg_zoomfov 20` and `set cg_fov 20` in console and try to zoom. Here the feel when zoomed in is noticeably slower that when zoomed out even though the field of view is exactly the same!

The short mention of a toggle bind is a workaround to get the optimal mouse sensitivity within the current gamecode. If you move away from using `+zoom` and instead use a toggle bind that changes `cg_fov` you can then adjust `m_yaw` and `m_pitch` to get the optimal feel according to the formula. This is however only possible when switching to a toggle bind so if you prefer to hold your zoom button for zooming there is no solution for you.

The toggle bind can look like this:

```
seta zoomin "set cg_fov 61.30133591; set m_pitch 0.009777778; set m_yaw 0.009777778; set zoom vstr zoomout"
seta zoomout "set cg_fov 106.2602047; set m_pitch 0.022000000; set m_yaw 0.022000000; set zoom vstr zoomin"
seta zoom vstr zoomin
bind mouse2 vstr zoom
```

Here `m_pitch` and `m_yaw` while zoomed in have been calculated using `0.022 * (tan(zoomfov / 2) / tan(basefov / 2))`.

What the change I have suggested does here does is to fix the scaling done by the game when `+zoom` is active. That would remove the need for ugly workarounds and would allow the players to use the optimal formula with hold-zoom.

No need to apologize.

The-Gig:
Okay, now I undestand the "toggle bind" thing, it's a workaround for fixing aiming sensibility during zooming with current gamecode (and the only possible way with any old mod, I guess), by using a "fake zoom". Instead of using the game's zoom feature, you temporarily change cg_fov and alter mouse movement speed accordingly, to prevent the mouse from moving too fast.

I just tried your script, it looks like it works. Maybe we may even write about this configuration trick somewhere in the wiki...

leilei-:
FWIW the engine's renderer is what's doing the hor+ stuff and shouldn't interfere witht he cg code in any way (other than reading the cgame's fov to calculate a new refdef view in tr_scene.c)

ryvnf:
@The-Gig This discussion is becoming a bit off topic of the change. If you have more questions not directly related to the change you can ask me on Discord or IRC.

The "mouse speed" slider in the menu controls the cvar `sensitivity`.

For the scaling to work properly it has to affect both mouse sensitivity and acceleration. Adjusting `m_yaw` and `m_pitch` by a factor will change mouse sensitivity and acceleration by that factor. You could also change `cl_mouseaccel` and `sensitivity` by the same factor instead. But I find it cleaner to adjust `m_yaw` and `m_pitch` because then I can freely change `sensitivity` and `cl_mouseaccel` and the zoom sensitivity will be updated accordingly.

@leilei- 

That is good news!

The-Gig:
ryvnf, thank you for your explanations. Added on Discord, so we may talk there if needed (I just need to remember that "commie" and "ryvnf" are the same person. ^_^).

The-Gig:
I and ryvnf created a wiki page about his mouse sensitivity scaling _trick_, which may still be useful even after this commit is accepted, due to working with any old mod. It also allows inserting additional levels of zoom. Of course, it has got its own drawbacks.
http://openarena.wikia.com/wiki/Configuration_examples/Scale_mouse_sensitivity_with_zoom
There is also an .ods spreadsheet which creates the "manual zoom" script, so you just need to copy-paste it into a configuration file.

The-Gig:
It looks like there might be some relationship with engine widescreen adaption formula after all, as ryvnf posted this related pull request in engine: https://github.com/OpenArena/engine/pull/37

ryvnf:
@The-Gig Yes, that was because old widescreen FOV-code doesn't scale the values as you would expect and is a bit flawed which breaks this gamecode change and the config tricks we wrote the wiki-page about.

But accepting that pull request will make the problem go away.

sago007:
The code looks good to me but I am not going to merge it if @leilei- says that we will be using a different approach. I am a bit confused about the conversation. The whole discussion means very little to me.

@leilei- is this something we want or not?

The-Gig:
I'll try to briefly sum up what I understood:
- When using zoom, you can feel a different mouse speed "feeling" than usual, especiallly if your cg_fov is different than 90. This is due to a too simplistic formula used by id software.
- He managed to apply a more accurate formula, which would be disabled by default (to keep the original behavior as default), but which players may enable by setting a new cvar to 1.
- In the meanwhile (or when using any old mod in the future), it is possible to use a configuration trick instead as workaround, and we made a wiki page about that trick.
- He found something strange in Leilei Hor+ formula, which breaks this fix with current nightly build binaries, and so prepared a fix also for engine.

The-Gig:
ryvnf, do you think there is any chance to apply a such fix in engine instead of gamecode, to make it also available for old mods?

(Note: I'm not 100% sure I haven't asked you this before, but doing a _quick search_ it seems I didn't.)

ryvnf:
It is probably possible to do the fix engine-side and have it be compatible with other mods using some hack. But that would be very ugly. The zoom sensitivity calculation takes place in gamecode, so it is the logical place to put the fix. And to make the fix-engine side the engine has to override the sensitivity value written by the gamecode, which could lead to other problems.

I also don't think that the engine knows when the player is zooming. Which would make it harder to do the fix there, as it doesn't know when to use the formula.

ryvnf:
I figured out a way to fix this in the engine code, the change makes the engine automatically scale the sensitivity according to the current FOV, completely ignoring the cgame calculations. This change also affects when players change the `cg_fov` setting, not only when zooming. This is probably what you want (being able to switch settings while preserving the same mouse feel), but it does change the base sensitivity for existing configs that have `cg_fov != 90`.

[See this commit for more information](https://github.com/OAGuild/yuengine/commit/a5e2ec266fdc90593e305ff4b46458c12f6016a8)

The-Gig:
Interesting...

Maybe @leilei- may tell us if she would prefer a such approach ("cl_scaleSensWithFov" -engine trick completely overriding cgame sensitivity if enabled, applies to zoom, fov changes in general and also to mods-) or your original idea ("cg_zoomSensitivity" -gamecode fix applied to zoom only and does not apply to mods).
I can guess both of them require the engine HOR+ formula to be fixed first...

ryvnf:
I'm closing this pull-request.As there was a better solution by changing to engine-code to solve the problem.

Ofc, you may use this change anyways, if you think it is better than the more general solution which applies to FOV change in general (not just when zooming). But I think it is better to solve the general case instead.

If there is interest I can make a pull-request to the engine which implements the other solution. Also note that the change will still be incompatible with the incorrect Hor+ formula currently used in OpenArena engine. That also breaks the mouse sensitivity tricks which me and Gig wrote about in the wiki. So I urge you to accept the pull-request to fix that bug, even if you don't agree on the other changes.

But as none of my pull requests has gotten any feedback from the developers, I'm not going to write the change to the engine-code unless explicitly requested. I'm a bit disappointed by the lack of response to my changes to be honest.

PR #20, opened 2017-08-22
vmMain(): add Q_EXPORT macro in definitions

GuildDio:
Without this, the following error happens when trying to run with
native libraries (.so):

```
Sys_LoadGameDll(/path/to/qagamex86_64.so) failed to find vmMain function:
"/path/to/qagamex86_64.so: undefined symbol: vmMain" !
Failed to load DLL /path/to/qagamex86_64.so.
----- Server Shutdown (Server fatal crashed: VM_Create on game failed) -----
```

### DIFF
--- a/code/cgame/cg_main.c
+++ b/code/cgame/cg_main.c
@@ -46,7 +46,7 @@ This is the only way control passes into the module.
 This must be the very first function compiled into the .q3vm file
 ================
  */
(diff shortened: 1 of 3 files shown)


sago007:
I just confirmed it, tested it and merged it. I guess that you where mainly interested in the "game" part but I also fixed the "cgame" and "UI" part (in a separate commit).

PR #21, opened 2017-10-16
Sorting entities by order for the editor. Also first pass on comments.

NeonKnightOA:


em92:
Have you tried to connect to server with your modifications from original client?
Server with your modifications will have ammo_belt with item index equal to 1. It will force client to load item with index 1. On original client item with index 1 is item_armor_shard. So I guess, if map doesn't have any item_armor_shards, client won't load ammo_belt.

If this happens, I recommend to make pull request in your editor to sort entities and close this one.

The-Gig:
Do you mean that items are given a specific number based upon their position in the source files, and thus changing the order of the items inside the files may be a bad idea, due to causing inconsistencies between client and server?

NeonKnightOA:
Reverted to previous order, then.

sago007:
I have been thinking a bit about. I am a bit worried about the high amount of repetition. I have an idea and hope to post a suggestion on https://github.com/NeonKnightOA/oagamepack on a different way of handling it later today. That will keep the very Editor specific options out of the code and open the possibility of translations.

NeonKnightOA:
As long as the _oa.game/baseoa_ folder keeps ready-to-be-used entities.def and entities.ent in there... remember that other repos read the gamepack repo in order to download the files.

PR #24, opened 2018-04-09
Made it compile by adding a forward declaration. 

sago007:
Also changed indentation slightly in some code that I tended to misread.

This is of course just my suggestion. I placed the trap_Cvar_VariableStringBuffer near the other trap functions where I believed that it made the most sense.


sago007:
This is redundant after Pull Request #40

PR #29, opened 2018-11-18
Better control of what's spawned on the map.

NeonKnightOA:


The-Gig:
Excuse me, what do you mean with this?

> // Double Domination and Domination don't support runes.

And about hiding red and blue flags (which, unlike some other gametype objective, in 0.8.8 aren't automatically hidden in gametypes where they are not used):
Maybe Sago had some reason to do not automatically hide also them, I don't know...
One possible reason may be that older gamecode (e.g. mods) would still show them, so maybe keeping mappers accostumed to placing "gametype" key on them may be good. Just guessing...

sago007:
@The-Gig My only reason for not hiding them was to not make it too different from vq3.

Regarding the code:
```
if(g_runes.integer)
    {
        if(item->giType == IT_PERSISTANT_POWERUP)
        {
            ent->s.eFlags |= EF_NODRAW;
            ent->r.svFlags |= SVF_NOCLIENT;
        }
    }
```
This code will never be reached because it drops out due to G_ItemDisabled(item)


The line 
```
if(g_gametype.integer != GT_CTF || g_gametype.integer != GT_1FCTF || g_gametype.integer != GT_CTF_ELIMINATION)
```
is always true! It should properly have been 
```
if(g_gametype.integer != GT_CTF && g_gametype.integer != GT_1FCTF && g_gametype.integer != GT_CTF_ELIMINATION)
```


Some of the lines are indented with spaces. The rest of the code uses tabs.

NeonKnightOA:
Applied the requested changes.

The-Gig:
@NeonKnightOA, why the code does still hide the runes in Domination?
For Double Domination okay, 0.8.8 already hides them, as I can guess Guard and Scout would be OverPower there, making scoring very difficult.
But _Domination is different_, so I don't think there is a reason to remove them there, unless the mapper or the server admin wants so. In 0.8.8 they are there.
Futhermore, I'm not sure, but maybe "hidden" items can still be touched and grabbed...

And about the "single player" mode, for how the game is now, I would say that it should be made the same as FFA regarding this stuff (otherwise, you may add a map to single player ladder and then discover there is useless stuff spawning there). But I don't know how the OA3 single player mode will actually work, considering someone mentioned a ladder with different game modes...

PS: The "quality review" auto control still reports a problem, it looks like a series of OR which will always return True, as it currently reads "IF (gametype not ctf OR gametype not 1fctf OR gametype not ctfe)" and _surely_ gametype "will not be" at least one of them. Not sure about the right C syntax for what you mean. Looking at other lines you wrote there, maybe just replacing the OR || with an AND &&?

NeonKnightOA:
Domination is a base-less gametype. It doesn't make sense to have team-locked runes.

The-Gig:
> 
> 
> Domination is a base-less gametype. It doesn't make sense to have team-locked runes.

You know that runes can be placed without team-lock, e.g. in the middle of the map. That's how I did in OACMPDM10, and IIRC you mentioned you also did it in one of your recent czest2ctf betas. _So, I think in this case the choice should be left to map author._
Now that I think about it, playing Domination in ctf-based maps with team-locked runes may still be an incentive for players to roam the map even far from the Domination point, to reach their team-dedicated runes.

Also, runes (team-locked or not) spawn in other "base-less" gametypes such as FFA and TDM, why should Domination differ?

NeonKnightOA:
In-game, IICR, team-locked runes are only available to the team that picks them. Blue team can't pick up red runes, and red team can't pick up blue runes. This is regardless of gametype.

The-Gig:
> 
> 
> In-game, IICR, team-locked runes are only available to the team that picks them. Blue team can't pick up red runes, and red team can't pick up blue runes. This is regardless of gametype.

Yes. But the fact you point this out makes me wonder if some parts of my previous post weren't clear. Should I reword something? Tell me, no problem! :-)
If that was about the _last phrase_ of my previous post, I'll try to explain it better:
If you are red, you can pick up "red-locked" runes and "team free" runes. If you are blue, you can pick up "blue-locked" runes and "team free" runes. But when you are playing in any non-team-based mode, you can pick up "team free" runes _only_, as you are not red and you are not blue... so maybe the game may directly hide red-locked and blue-locked runes in those gametypes, showing only those which are "team free". In team-based modes, all three kinds of runes should be shown instead.

NeonKnightOA:
Because Domination isn't as played as FFA, TDM, CTF or Elimination. Because the game and the serveradmins don't incentive people to go outside of the popular modes. And the fact that serveradmins don't complain about the problem doesn't mean that the problem is not there.

In Domination, areas are supposed to be neutral for both teams. I don't think I have to repeat that again nor anything I've said before, Gig. It's NOT a good idea, it will NEVER be. What could be implemented is the nullification of both RED_TEAM and BLUE_TEAM spawnflags for non-base-based modes so runes can be picked up by anyone.

sago007:
Could you give an example on how to test your modifications? I seem to have a bit of trouble seeing the difference on the maps I have.

Regarding the code:
Still space indented. Tabs are dominant in the file.
You are using Allman style braces. K&R is dominant in the file.
Some comments seem redundant. "Runes are spawned with a specific configuration." tells me nothing. A comment like "Spawn flags like RED_TEAM or BLUE_TEAM must be written to generic1"

The-Gig:
> Could you give an example on how to test your modifications? I seem to have a bit of trouble seeing the difference on the maps I have.

Waiting for NK's answer, I think -apart the "Dom+Runes" thing which I'm still not persuaded- (about it, without furhter changes elsewere, isn't there the risk they would still be picked up despite being hidden?), the other changes simply check that items are actually hidden in the gametypes where they are useless. Something that for some items already happens (e.g. obelisks) and for others not (red and blue flags).
So, I guess the way to test it would be to make a small test map with _all_ "gametype objective" entities inside (plus the runes), without any "gametype" or "!gametype" key, and open it in all gametypes, to be sure about what actually spawns there and what not.

NeonKnightOA:
And I'm not a fan of giving advantages to teams in team-neutral gametypes such as Domination. It just doesn't make sense. You're turning the balance of the map artificially in favor of one team.

Anyway, branch reboot, closing the PR.

PR #32, opened 2018-11-29
Missionpack checks

NeonKnightOA:
I had disabled these pieces of code in order to test important stuff for the SP. However, instead of commenting them out, I thought it would be better, in case team models are going to be implemented in a future (the idea itself is good, I can see clans creating their own model with exchangeable heads), to just create a toggleable cvar.

Also some pieces of commented code weren't parsed well in some editors, so I've changed the commenting format.

### DIFF
--- a/code/cgame/cg_main.c
+++ b/code/cgame/cg_main.c
@@ -298,10 +298,6 @@ vmCvar_t cg_weaponBarStyle;
 vmCvar_t cg_chatBeep;
 vmCvar_t cg_teamChatBeep;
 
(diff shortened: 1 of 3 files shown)


NeonKnightOA:
I'll wait for Sago's decision about #35 before continuing to work on this, as there are conflicts between that request and this one.

NeonKnightOA:
Code updated.

sago007:
Added in https://github.com/OpenArena/gamecode/commit/b6ff88cee0db98900d56e3e03f45be7eabb5c091

PR #33, opened 2018-11-29
New cvar for Harvester: g_harvesterFromBodies.

NeonKnightOA:
This cvar allows Harvester to work in a similar way to Unreal Tournament 3's Greed mode: it eliminates the neutral obelisk and makes the skulls appear from the fallen players' bodies instead.

Only one known issue, and is that the neutral obelisk, in spite of the checks, is still drawn in the map.

### DIFF
--- a/code/cgame/cg_consolecmds.c
+++ b/code/cgame/cg_consolecmds.c
@@ -400,7 +400,7 @@ static void CG_DenyOrder_f(void) {
 }
 
 static void CG_TaskOffense_f(void) {
-	if (cgs.gametype == GT_CTF || cgs.gametype == GT_CTF_ELIMINATION || cgs.gametype == GT_1FCTF) {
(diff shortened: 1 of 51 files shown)


sago007:
You might want to keep the neutral obelisk in case of a skull spawning in a nodrop zone. A skull dropped in a nodrop zone should be spawned at the neutral obelisk or the team obelisk furthest away from the target if the neutral obelisk is unavailable.

NeonKnightOA:
An alternative path could be that in case of falling to the nodrop zone, the skull doesn't spawn. I'll see about that.

NeonKnightOA:
Code updated. The cvar was also renamed so Gig doesn't complain.

NeonKnightOA:
Since this was added, it's only appropriate to close the PR and delete the branch.

PR #34, opened 2018-11-29
Grapple support: g_grapple.

NeonKnightOA:
**g_grapple** makes players start with the Grappling Hook in their inventory.
~**g_grappleAll** makes players start with the Grappling Hook in Instagib and All Rockets modes; it requires g_grapple != 0.~

### DIFF
--- a/code/cgame/cg_consolecmds.c
+++ b/code/cgame/cg_consolecmds.c
@@ -400,7 +400,7 @@ static void CG_DenyOrder_f(void) {
 }
 
 static void CG_TaskOffense_f(void) {
-	if (cgs.gametype == GT_CTF || cgs.gametype == GT_CTF_ELIMINATION || cgs.gametype == GT_1FCTF) {
+	if (CG_UsesTeamFlags(cgs.gametype)) {
 		trap_SendConsoleCommand(va("cmd vsay_team %s\n", VOICECHAT_ONGETFLAG));
 	} else {
 		trap_SendConsoleCommand(va("cmd vsay_team %s\n", VOICECHAT_ONOFFENSE));
--- a/code/cgame/cg_draw.c
+++ b/code/cgame/cg_draw.c
@@ -1443,7 +1443,7 @@ static float CG_DrawFollowMessage(float y) {
 	char *s;
 	int w;
 
-	if (!(cg.snap->ps.pm_flags & PMF_FOLLOW) || ((cgs.elimflags & EF_NO_FREESPEC) && (cgs.gametype == GT_ELIMINATION || cgs.gametype == GT_CTF_ELIMINATION))) {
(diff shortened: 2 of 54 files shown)


sago007:
Is it necessary with g_grappleAll? Could g_grapple not cover both? Why does they have different cvar flags?

NeonKnightOA:
I thought elimination_grapple was restricted to only Elimination. Haven't checked that yet.

grappleAll exists if players want to play with the grapple, but not in Instagib and All Rockets mode.
As for the different cvar flags, I thought the server should only send that it supports grapple, as the modes it's supported are irrelevant, I guess. But yeah, I now have second thoughts about grappleAll.

The-Gig:
Yes, elimination_grapple works in Elimination/ECTF/LMS modes only (and in any other mode in case g_elimination is enabled). IIRC, what it does is to give the grapple to all players at spawn.
So, if there will be the new g_grapple which will do the same thing (giving grapple to all players) in all modes, it looks like elimination_grapple would be reduntant then (so it may be deleted to save a bit of precious cvar space).

Simple brainstorming: maybe one may even foresee g_grapple -1 and/or g_grapple 2, to also prevent the grapples placed in the map from spawning (in standard/grapple modes). However it's already possible to do that through _set disable_weapon_grapplinghook 1_, so maybe it would be reduntant.
Just thinking about it... neat values could be: 0=no grapples at all, 1=only grapples in the map (default), 2=players spawn with it (so they may be removed from the map).

The-Gig:
I just noticed your
> As for the different cvar flags, I thought the server should only send that it supports grapple,
so I gave a look and so I guess this line
> { &g_grapple, "g_grapple", "0", CVAR_SERVERINFO | CVAR_NORESTART, 0, qfalse},
means that g_grapple is set as serverifo (like a cvar set using "sets") mode. I fear that _only **really, really** important cvars should be set as serverinfo by default_, because there are tight limits for the lenght of the serverinfo strings: exceeding it causes _major_ problems:
Quote about "sets" command from a certain wiki:
> It may be very useful, but the only problem is that there is a limit on the data that can be passed through this method: if you see an "Info string length exceeded" message in your server console, this means that you have set too many variables as "sets", so you to fix that you either have to set less variables as "sets", or you should shorten them (the data contained inside the strings). After you see it, if you make your init string even longer, adding more or longer "sets" variables, your server may even become inaccessible for players. 

I think in the past days you also created some new cvars... please check that you didn't set them as "serverinfo" unless that's really, really necessary!

NeonKnightOA:
I'll wait for Sago's decision about #35 before continuing to work on this, as there are conflicts between that request and this one.

NeonKnightOA:
I've just checked elimination_grapple, it almost doesn't appear in the code outside of the cvar definition sections. I don't think it's important to delete it.

sago007:
Added to master on https://github.com/OpenArena/gamecode/commit/cbaf971179581b9841315e9a1905b81d5c15f450. I also eliminated elimination_grapple and optimized the code. The code growth was limited to 7 lines.

PR #35, opened 2018-11-29
Teamgame checks moved to centralized functions.

NeonKnightOA:
It got bothersome to check if a gametype is team-based or not, and led to errors. So I've created 4 functions, one at cgame, one at game, one at q3_ui, and another at mp_ui, to have a centralized place where to make general teamgame checks.

### DIFF
--- a/code/cgame/cg_consolecmds.c
+++ b/code/cgame/cg_consolecmds.c
@@ -400,7 +400,7 @@ static void CG_DenyOrder_f(void) {
 }
 
 static void CG_TaskOffense_f(void) {
-	if (cgs.gametype == GT_CTF || cgs.gametype == GT_CTF_ELIMINATION || cgs.gametype == GT_1FCTF) {
+	if (CG_UsesTeamFlags(cgs.gametype)) {
 		trap_SendConsoleCommand(va("cmd vsay_team %s\n", VOICECHAT_ONGETFLAG));
 	} else {
 		trap_SendConsoleCommand(va("cmd vsay_team %s\n", VOICECHAT_ONOFFENSE));
--- a/code/cgame/cg_draw.c
+++ b/code/cgame/cg_draw.c
@@ -3104,7 +3104,7 @@ static qboolean CG_DrawScoreboard(void) {
 #else
 	char *s;
 	int w;
-	if (cg.respawnTime && cg.snap->ps.persistant[PERS_TEAM] != TEAM_SPECTATOR && (!CG_IsARoundBasedGametype(cgs.gametype)) {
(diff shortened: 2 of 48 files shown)


sago007:
Could you explain what "CG_IsATeamGame" does?
The name *strongly* suggests that it returns true if the gametype is a team game and false otherwise.
The capturebased parameter does not make any sense either.

Because this has annoyed my many times, I believe that I know what you are trying to do.

I think you want to split CG_IsATeamGame into two:
qboolean CG_IsATeamGame(int gametype)
qboolean CG_UseBaseSpawns(int gametype)
I think this covers most scenarios.

I believe only TDM is a bit inconsistent being the only team game that uses "fraglimit". In the UI, it can be called "score limit" for all gametypes to save some logic.

NeonKnightOA:
The check I wanted to do is as follows:

Does the gametype use a capturelimit?
- If passed the checklist, return true, if not, return false.
Else (the gametype use fraglimit)
- If passed the checklist, return true, if not, return false.

Why the division? Because of TDM of course, as it's the only team-based gametype that uses fraglimit (a case could be made for Elimination and LMS as well, but those use capturelimit instead).
My idea is to make it as generic as possible so future gametypes and mods of OA could take advantage of it.

sago007:
I do not understand it.

Be adding the boolean you make it less generic and hard to read.

To really share code you may consider some C macros:
```
#define GAMETYPE_IS_TEAM_GAME(gametype) (gametype>= GT_TEAM && gametype != GT_LMS && gametype != GT_GT_POSSESSION)  
#define GAMETYPE_USES_CAPTURELIMIT(gametype) (gametype>GT_TEAM && gametype != GT_GT_POSSESSION)  
#define GAMETYPE_USES_BASES(gametype) (gametype>GT_TEAM && gametype != GT_LMS && g_gametype != GT_DOM && gametype != GT_GT_POSSESSION)
 ```

I have not tested the macros above and they may have logic errors.

Place them in a header file and include in all modules.


I also git reminded of this: https://medium.com/@amlcurran/clean-code-the-curse-of-a-boolean-parameter-c237a830b7a3

NeonKnightOA:
There, I think this step is done. I'll check that about the defines.

sago007:
Don't worry about defines. They are usually discouraged anyway.

NeonKnightOA:
I think I've tackled all of the checks.

There's one on bg_misc.c that makes the compiler whine.

sago007:
Squashed and commited to master on 0bee41a517f75ba62ed2bfb98f1c54d2fb5533a6

NeonKnightOA:
Thank you! I'll update the other PRs with the new code.

PR #36, opened 2018-11-29
Prevent key game objectives from being disabled.

NeonKnightOA:
Key game obejctives shouldn't be disabled because they break gamemodes.

If players want to screw the game for themselves, that's fine. But intentionally screwing other people's playing experience IS BAD.

### DIFF
--- a/code/game/g_items.c
+++ b/code/game/g_items.c
@@ -960,6 +960,14 @@ G_ItemDisabled
 int G_ItemDisabled( gitem_t *item )
 {
 	char name[128];
+	
+	/* Neon_Knight: Disable shouldn't work on key game objectives in online games.
+		You just can't trust people with always do the right thing. And OA shouldn't
+		suffer for that. */
+	if (item->giType == IT_TEAM) {
+		return qfalse;
+	}
+	/* /Neon_Knight */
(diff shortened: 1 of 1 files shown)


sago007:
I disagree on this change.
If the user wants to break the game, he should be allowed to do so.

NeonKnightOA:
Well, if someone wants to break a game, they should do so to screw only their own experience, not to screw everyone else's online experience. And disabling key game objectives in online games IS screwing everyone else's playing experience.

The-Gig:
Probably it's not a real issue... who would just use it? In OA 0.8.8 disable_ may be used to remove red/blue flags in gamtypes where there are not needed... otherwise I find it quite unlikely someone will disable key objectives.
Server admins probably have many more ways to break the game..

NeonKnightOA:
It's not a matter of "who would use it". A rule of thumb is "everyone can do bad things" even if they don't. You can't blindly trust people to do the right thing, Gig.

For key objective hiding there's already the spawn functions in the gamecode.

By using disable_ on key items without the need to do so IN AN ONLINE SERVER, the serveradmin is ruining the experience for the players who go there.

NeonKnightOA:
This has been stale for a long time and there doesn't seem to be any intention of fixing the issue. Closing it.

PR #37, opened 2018-12-02
[Testing required] droprune: Command to drop the currently held rune.

NeonKnightOA:
This is a command that, baffingly, was absent from TA. I say baffingly because the equivalents (the Runes for the Threewave CTF mods for Q1 and Q2 and Unreal Tournament's Relics) already had keybinds for dropping the currently held rune.

So this keybind does what it says: it drops the rune back to its spawning point.

**[Here's a test pk3.](https://cdn.discordapp.com/attachments/405518544382590987/675451968453476353/z_oax_v2.pk3) (Updated Feb 7, 2020 18:37)**

### DIFF
--- a/code/game/g_cmds.c
+++ b/code/game/g_cmds.c
@@ -662,7 +662,14 @@ void Cmd_LevelShot_f( gentity_t *ent ) {
 	trap_SendServerCommand( ent-g_entities, "clientLevelShot" );
 }
 
-
+void Cmd_DropRune_f( gentity_t *ent ) {
+	if (!ent) {
+		return;
+	}
+	if (g_runes.integer >= G_RUNES_ENABLE_TOSS) {
+		TossClientPersistantPowerups(ent);
+	}
+}
 /*
 ==================
 Cmd_LevelShot_f
@@ -2310,8 +2317,11 @@ commands_t cmds[ ] =
 	//KK-OAX
 	{ "freespectator", CMD_NOTEAM, StopFollowing },
 	{ "getmappage", 0, Cmd_GetMappage_f },
-	{ "gc", 0, Cmd_GameCommand_f }
-
+	{ "gc", 0, Cmd_GameCommand_f },
+	
+	/* Neon_Knight: Toss current rune*/
+	{ "droprune", 0, Cmd_DropRune_f },
+	/* /Neon_Knight */
 };
 
 static int numCmds = sizeof( cmds ) / sizeof( cmds[ 0 ] );
--- a/code/game/g_local.h
+++ b/code/game/g_local.h
@@ -1432,3 +1432,8 @@ qboolean G_UsesTeamFlags(int check);	/* Whether the gametype uses the red and bl
 qboolean G_UsesTheWhiteFlag(int check);	/* Whether the gametype uses the neutral flag. */
 qboolean G_IsARoundBasedGametype(int check);	/* Whether the gametype uses the neutral flag. */
 /* /Neon_Knight */
+
+/* Neon_Knight: Needed for g_runes */
+#define	G_RUNES_ENABLE		1
+#define	G_RUNES_ENABLE_TOSS	2
+/* /Neon_Knight */


sago007:
I have always been reluctant to include changes that required key binds.
In my view the low amount of key binds is a major strength of the game.
I fear that more key binds will increase the entry barrier to the game. For a mod this is not a problem  but for a game that you want to be able to introduce to new players it is.

NeonKnightOA:
This shouldn't be a big issue. It's just adding something that should be there since day one. I don't think is that gamechanging. The other solution would be adding "Suicide" to the keybind menu, and I think that's a bit too much.

The-Gig:
Usage of "kill" binding is a bit controversial among players (some think it's perfectly legit, others think it's unfair). However the command always existed so I don't think adding it to menus would be so bad after all. Maybe one may foresee a server cvar to enable a delay of a few seconds before the suicide is actually performed.

About dropping runes M2C are that should be allowed or disallowed by the server, if we want to add it. Just M2C.

NeonKnightOA:
"kill" makes a player respawn at the cost of losing its inventory. What should be implemented in that case is a delay so it doesn't get as much abused.

"dropweapon" (another keybind I plan on implement) is also a commonly used keybind in a lot of games in order to share weapons between teammates. Hell, the first two Quakes had it IICR.

"droprune", as I've explained (and I hope not to have to repeat this again, please) has also a bit of history, in that it allows players to drop a rune for another teammate who may want to have it.

None of these are big game changers.

The-Gig:
Then maybe you may create a bitfield variable that, with different values, may allow to drop weapons, powerups (which one, in case of having more? All of them?), runes, armor (yellor armors... in case you have >50... or just shards?). Of course, disabled by default. Just a random idea... I'm NOT telling it's a good idea. I still think OA isn't PUBG or Fortnite...

NeonKnightOA:
I've opened a thread on the forum. If lei says this is NOTTODO, then I'll close the PR.

The-Gig:
"Dropping" here a link to the forum thread: http://openarena.ws/board/index.php?topic=5431.0 :)

The-Gig:
While I still think that this should be server-controlled and disabled by default (e.g. enabled by a DMFLAGS), if you can provide a test pk3, I would give it a try.

NeonKnightOA:
[Here it goes](https://cdn.discordapp.com/attachments/405518544382590987/675339088877846548/z_oax_v2.pk3)

The-Gig:
I gave it a quick try, it seems to work. You press the key, and the rune respawns in its spawn point (and you don't have it anymore, of course).

The-Gig:
I was thinking... if the runes are the *only* thing meant to be dropped, maybe the most efficient thing would be *g_runes 2* to allow it.

If instead *more things* are planned to be dropped, then a cvar like *g_dropItems* may fit. Maybe even as a bitflag (or however multivalue) to customize *what* can be dropped, if you like so...

NeonKnightOA:
Well, yeah, the name of the bind says so. "droprune" is meant to only throw runes.

NeonKnightOA:
[Updated test pk3](https://cdn.discordapp.com/attachments/405518544382590987/675451968453476353/z_oax_v2.pk3). Introduces `g_runes 2` which enables rune tossing.

I'm not really particularly fond of this idea (in fact, I would prefer to just delete g_runes entirely and have runes enabled by default, as there are maps that rely on them), but it should be put to test to see if it really works.

The-Gig:
I'm sorry I can't try the new pk3 right now, I'm from cell phone with my baby sleeping in my arms.

However, a few points:
- You mentioned also a "dropweapon" idea... So I wanted to say that we might plan things carefully to avoid getting N different cvars to allow/disallow dropping of N different item kinds, as cvars space is limited I think. (On a side note, *I don't know* how much useful a "dropweapon" key is in a game where weapons respawn every 5 seconds and you can strafejump the whole map in less than 20 seconds...)
- Maybe we may discuss (trying to involve the community as possible) about changing the default value of g_runes, but I think the cvar is useful and I suggest to keep it. I know it's possible to alternatively use "disable items" feature, but that's more complicated and unknown to the average user I guess.
- How do you plan to make bots use the droprune feature? Somehow make them decide they prefer another rune than the one they are currently carrying, so they drop the old one to get the new one? A new team order to tell your teammate "Grism drop rune (so I can grab it)"?

NeonKnightOA:
Team Deathmatch has a default respawn time of 30 ever since Q3A itself. That's one gametype that could be benefitted from dropweapon.

As for the bots, that would be an order, and upon giving it to the bot they may reply if they have a rune or not. If they have a rune, they may drop it, if not, they would report they don't have a rune. As for how likely the bot may drop the rune, that would depend on how much they value other items:

* If they have a high affinity for Haste and tasks where speed is fundamental (i.e rushing to base, taking the flag and escaping with it), they would be inclined to pick up and keep the Scout rune.
* If they have a high affinity for health, armor, Regeneration and defense-oriented tasks, they would be inclined to pick up and keep the Guard rune.
* If they have a high affinity for explosive weapons, Quad Damage and damage-dishing-oriented tasks, they would be inclined to pick up and keep the Doubler rune.
* And if they have a high affinity for non-explosive weapons and all the other items, as well as for any other kind of tasks, then they would be inclined to pick up and keep the Ammo Regen rune.

If the bot has an affinity for the rune they're carrying, it would be half a chance that they would drop it. Likewise, if they don't have an affinity for such rune, they would be quick to drop it.

**However, by now, I would prefer to focus in this particular branch for the basics to work.**

NeonKnightOA:
In the end I've decided to go with a non-bitflag cvar. Why? Because bitflags only make sense if all of its combinations make sense (it's why it was fitting for devmode, where you may want to enable none, one, few, many or all of its modes). And in this case the combination "disabled (0) + allow rune throw (2)" didn't made sense.

The-Gig:
I agree. A bitflag cvar would only have made sense for something like g_dropItems or g_allowDrop with (example) 1=flags, 2=runes, 3=powerups, 4=holdables, 5=weapons, 6=armor...
But I don't know if I want a game with so many key bindings...

NeonKnightOA:
I think only runes and weapons make sense for a game like OA. Powerups are temporary, Holdables IDK, and armor and health? That's never seen before outside of class-based FPS.

The-Gig:
If I would have to choose only one kind of item to drop, I think I would say "flags". Because that would be a "mutator" which would be noticeable change from the classic (should make capturing easier for who can do a little of teamwork) and easy to learn ("I am low on health, then I pass the flag to a.teammate")....

The-Gig:
Tried the new pk3. It seems to work very well! 👍 

Now, I was thinking a few questions (just asking):
- Do you plan to add some kind of feedback when dropping a rune? Like playing a sound (e.g. the same of when you pick up a rune?) or showing a "rune dropped" message somewhere?
- Do you think it may be useful to write something like "Rune dropping is not enabled on this server" in console (similar to "cheats are not enabled on this server") in case one has got a rune and presses the button, but g_runes is <2? I don't know.

PS: In my previous message I forgot to type the initial part of the phrase: "You know I don't think dropping anything is really "necessary" in this game, but if I would have to choose only one..."

NeonKnightOA:
Closing since this isn't going to be considered.

The-Gig:
I hope you may reopen it later, as it was interesting & working...

PR #38, opened 2018-12-02
Check for a starter weapon: replaces searches for starting weapons.

NeonKnightOA:
Same deal as the gametype checks. These checks verify if a weapon is a starter weapon (Gauntlet, Machinegun or Grappling Hook).


sago007:
This does not seem to lower the complexity. What if I play a gametype where I pick the machinegun or grappling hook?

Some of the checks are highly misleading. Why does the fact that it is a starter weapon affect the muzzle flash?

PR #39, opened 2018-12-02
Random number generator for UI3.

NeonKnightOA:
This function outputs a random number which can later be used in UI3.
It can then be used for a lot of different features such as randomized Arcade Modes, Quick Matches, and other gamemodes where you need the features to be randomly generated.


sago007:
I think RAND_MAX (which for some reasons is not defined in the bg_lib.h) is 0x7fff (or 32767). If you don't mind then I'll take a look on doing something like:
```
ui_randomIntToCvar CVARNAME minInt maxInt
ui_randomFloatToCvar CVARNAME minFloat maxFloat
ui_randomStringToCvar CVARNAME string1 string2 string3 ...
```
I think you can make more fun with those.
I'll try to take a crack at it Wednesday.

NeonKnightOA:
Yes! No problem! My intention with this is to make use of it in UI3 for certain SP features.

NeonKnightOA:
Definitely. That was better than my idea, even!

Closing. Thank you, Sago!

PR #44, opened 2018-12-08
Added missing yet declared functions to check for controlpoint-based gametypes and fixed a logic error in cg_newdraw.c.

NeonKnightOA:


sago007:
I gave added the Possession check.
I removed GAMETYPE_USES_CONTROL_POINTS from the original pull because I discovered that the few places where it was used should not have been there in the first place.

NeonKnightOA:
Ok. Closing.

PR #45, opened 2018-12-08
New cvars for Double Domination: g_ddCaptureTime and g_ddRespawnDelay

NeonKnightOA:
Branch had to be redone. I'm still trying to get used to git rebases. -.-

Anyway. These two cvars control **the amount of time required to score a point** and **the amount of time between rounds**.

There's a problem, and it's that the capture time isn't properly displayed in the screen. I tried everything to no avail. Also it would be a good idea if the time before a round starts is also displayed in a big message, and if a message can be displayed when a control point is taken by a team, like UT200X does.

### DIFF
--- a/code/game/g_bot.c
+++ b/code/game/g_bot.c
@@ -419,7 +419,7 @@ void G_CheckMinimumPlayers( void ) {
 		return; //If no AAS then don't even try
 	}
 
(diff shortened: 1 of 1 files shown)


sago007:
I can try a go at it. I think I know what is needed.

PR #46, opened 2018-12-08
Botfixes and new bot-related cvars

NeonKnightOA:
Resume of the branch:

- New cvar: **bot_allowTeamChat**, allows the bots to output regular FFA chatlines in team games. The cvar is disabled in Developer mode.
- Improvements to the AI in Domination, Possession, Double Domination and other gametypes so they play more as intended.
- Developer mode now uses the bots to output helpful messages instead of regular botchat.

### DIFF
--- a/code/game/ai_chat.c
+++ b/code/game/ai_chat.c
@@ -384,10 +384,18 @@ int BotChat_EnterGame(bot_state_t *bs) {
 
 	if (bot_nochat.integer) return qfalse;
 	if (bs->lastchat_time > FloatTime() - TIME_BETWEENCHATTING) return qfalse;
-	//don't chat in teamplay
-	if (TeamPlayIsOn()) return qfalse;
-	// don't chat in tournament mode
-	if (gametype == GT_TOURNAMENT) return qfalse;
+	//don't chat in teamplay if not allowed
+	if (TeamPlayIsOn() && !bot_allowTeamChat.integer){
+		if(!trap_Cvar_VariableIntegerValue("developer")) {
+			return qfalse;
+		}
+	}
+	// don't chat in tournament mode if not allowed
+	if (gametype == GT_TOURNAMENT && !bot_allowTeamChat.integer){
+		if(!trap_Cvar_VariableIntegerValue("developer")) {
+			return qfalse;
+		}
+	}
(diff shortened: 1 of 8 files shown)


The-Gig:
Wasn's there already a cvar like g_botchat, g_chatbot or something similar? I'm from phone, I can't check right now. How does that differ from your new cvar? Wondering if your new feature may just be an additional value for that existing cvar instead...

sago007:
I find the name a bit confusing. Perhaps "bot_allowTeamChat" would be more accurate?
Also for logical booleans you should use:
```
if (TeamPlayIsOn() && !bot_allowTeamChat.integer) {
```
and
```
if(trap_Cvar_VariableIntegerValue("developer")){
```
So that you can read "if team play and not allow team chat then" and "if developer then"

Also: Why does developer mode have any influence on bot_freeChat?

The-Gig:
Please confirm if I understood correctly:
- In non-team-based modes, bots are used to taunt.
- In team-based modes, bots are only used to talk about team orders, and they do not taunt there.
- bot_nochat, if set to 1 (default 0) prevents bots from taunting in non-team-based modes. They still talk about team orders in team-based modes.
- You new cvar (bot_freeChat) would allow or disallow bots to taunt in team-based modes.
Is that correct?

Uhm... maybe that may be done with something like **bot_nochat -1**, instead? Or alternatively, something "clearer"? _bot_tauntInTeam_ or similar?

PS: I thought developer mode only added additional debugging infos, I didn't think it would disable some kinds of standard outputs...

The-Gig:
... or maybe transforming **bot_nochat** into **bot_chat** (0=no taunt, 1=taunt in non-team-based only, 2=taunt in all modes). Default may be 1, I guess.
Just an idea, of course...

NeonKnightOA:
I'm not touching this anymore. I don't want to screw it up more than what's already done.

NeonKnightOA:
I've divided the three things I did in this branch into separate commits and updated to the current code.

sago007:
Wouldn't it be simpler to just set "bot_allowTeamChat" to 0 if developer is on?
There are a lot of trap_-calls, you should properly declare it and use a variable instead.

NeonKnightOA:
I think I just got an idea. Freezing this.

PR #51, opened 2018-12-09
[Testing required] Solving the problems with Shuffle

NeonKnightOA:
EDIT 2023-03-10 18:39 GMT-3
[Testing pk3 for Classic UI available](https://www.dropbox.com/s/qb1azflikd7vpeq/z_shuffle_test_oax.pk3?dl=0)

Shuffle was removed in another commit. According to Sago, the feature was removed because it can overflow the network stack and cause the game to crash if used online. Since this was a highly requested feature, I'm opening this PR so we can find a proper solution to it instead of removing it entirely.

The code contains a solution lifted from the Aftershock mod in a separate commit. It doesn't differ much from the current solution, TBH. But it could be a starting point to see what the problem might be with the function.

### DIFF
--- a/code/game/g_admin.c
+++ b/code/game/g_admin.c
@@ -194,6 +194,12 @@ g_admin_cmd_t g_admin_cmds[ ] = {
 		"display a (partial) list of active bans",
 		"(^5start at ban#^7) (^5name|IP^7)"
 	},
+	//KK-OAX
+	{
+		"shuffle", G_admin_shuffle, "f",
+		"Shuffles the teams"
+		""
+	},
 	{
 		"slap", G_admin_slap, "S",
 		"Reduces the health of the selected player by the damage specified",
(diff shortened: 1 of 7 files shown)


sago007:
Another solution may be to change the engine to be able to handle it. I know a new protocol was introduced in ioquake3 a long time ago. The original netcode includes obfuscation which makes no sense for a game with source code. I do not know if it fixes the problem.

The-Gig:
A question related with Suffle, although probably not related with its crashes:
What's the reason for not shuffling also bots, at the same time of humans? The problems that bots sometimes have recognizing their target changed team (continuing attacking an ex-opponent who become teammate, for some time) or something else?

sago007:
@NeonKnightOA I notice that you have worked on some things on this. How would you consider the status at the moment?

NeonKnightOA:
Can't check at the moment. My Linux HD, where I had most of my work, went belly up. I'm trying to fix it.

NeonKnightOA:
I need to rebase the branch to the new code. I haven't really worked on it, just ported the Aftershock mod solution in order to compare it with what we have. I think this is better solved by other people.

NeonKnightOA:
Testing pk3 available!

Closing in order not to clutter the PR list.

PR #54, opened 2018-12-19
cgame and game backports from ioq3 including tons of fixes and adapted backports.

NeonKnightOA:
The backports are mentioned at each commit, so no need to repeat them here.


The-Gig:
Is that "Improve finding obelisk entitynum for bot AI" thing related with those bot problems which may happen if the obelisk isn't perfectly aligned with the ground?

sago007:
i disabled the ioquake3 version of clientkick/kick. They went in a different direction.

PR #55, opened 2018-12-20
New cvar: cg_drawMinedMessageInstantly. Fuses into a single message t…

NeonKnightOA:
…he "You have been mined" and "Internal combustion" messages for players hit with a proxy mine.

### DIFF
--- a/code/cgame/cg_draw.c
+++ b/code/cgame/cg_draw.c
@@ -3285,10 +3285,8 @@ static void CG_DrawProxWarning( void ) {
 
 	proxTick = 10 - ((cg.time - proxTime) / 1000);
 
-	if (proxTick > 0 && proxTick <= 5) {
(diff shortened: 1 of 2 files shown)


The-Gig:
Uh? Does that need a new CVAR?

NeonKnightOA:
I would have replaced the old message for the new one, but there's always "that person". So I've made it a switchable cvar.

The-Gig:
I think almost every way is okay for me:
- Show one message first, then the other one
- Show them together, on two lines (as long as that does not interefere with something else in the HUD, e.g. frag messages)
- Show them together, on a single line (as long as it still fits in the screen also in 4:3)

For me, do as you wish... it's just that space for cvars is limited and maybe in this case a switch wasn't necessary, also considering that throwing mines directly at people it not their main use...
Do as you (and Sago, of course!) think it's better...

NeonKnightOA:
NVM, I'll just modify the message without the cvar. You're both right.

PR #59, opened 2019-01-29
UI (Q3/MP): Typo fixes, unification of gametype names, addition of Possession to the gamenames.

NeonKnightOA:


sago007:
I don't think you should change the classic ui names, the old UI is limited to 8 chars and it does not look good at the moment. I'll add Possession in a separate commit.

Also: I would recommend the name "Deathmatch" over "Free for all". vq3 only had one FFA mode but OA also have LMS and Possesion making "FFA" ambiguous.

PR #61, opened 2019-05-30
Isolated AI fixes and improvements for Possession and Domination

NeonKnightOA:
Sorry for the amount of commits, but I had to modify files from the web version of Github.

These changes aim to improve offline botplay for the Domination and Possession gamemodes. In Domination, bots are now willing to take the control points. As for Possession, they prioritize taking the flag and chasing the flag carrier.


sago007:
I tested this out.
Domination seemed to work fine.
I started a Possession game on pul1-dual. It did not work quite as good. While the bots were very aggressive while having the flag carrier in sight they were just standing still most of the time.

PR #62, opened 2019-05-30
AI fixes and improvements for Possession, Domination, Rocket jumping and Grappling

NeonKnightOA:
Redoing the pull-request again. Six months without coding practice leave quite a bit of rust. :/

Anyway. The commits are self-explanatory. These changes aim to improve offline botplay for the Domination and Possession gamemodes.

In Domination, bots are now willing to take the control points, and as a result, matches are more difficult than ever.

As for Possession, they prioritize taking the flag and chasing the flag carrier. Hopefully they also go to get the flag as well.

In addition, there's a new restriction for rocketjumping that checks if the bots can use the grapple and have it in their inventory. If they have the grapple and can use it (the safer alternative to rocket jumping) they won't use rocket jumps.

### DIFF
--- a/code/game/ai_chat.c
+++ b/code/game/ai_chat.c
@@ -380,20 +380,19 @@ BotChat_EnterGame
 */
 int BotChat_EnterGame(bot_state_t *bs) {
 	char name[32];
-	float rnd;
 
-	if (bot_nochat.integer) return qfalse;
-	if (bs->lastchat_time > FloatTime() - TIME_BETWEENCHATTING) return qfalse;
-	//don't chat in teamplay
-	if (TeamPlayIsOn()) return qfalse;
-	// don't chat in tournament mode
-	if (gametype == GT_TOURNAMENT) return qfalse;
-	rnd = trap_Characteristic_BFloat(bs->character, CHARACTERISTIC_CHAT_ENTEREXITGAME, 0, 1);
-	if (!bot_fastchat.integer) {
-		if (random() > rnd) return qfalse;
+	if (BotCanChat(bs,CHARACTERISTIC_CHAT_ENTEREXITGAME)) {
+		return qfalse;
 	}
-	if (BotNumActivePlayers() <= 1) return qfalse;
-	if (!BotValidChatPosition(bs)) return qfalse;
+	if (G_IsATeamGametype(gametype)) {
+		return qfalse;
+	}
+	if (BotCanChatInTournament()) {
+		return qfalse;
+	}
+	if (!BotValidChatPosition(bs)) {
+		return qfalse;
+	}
 	BotAI_BotInitialChat(bs, "game_enter",
 				EasyClientName(bs->client, name, 32),	// 0
 				BotRandomOpponentName(bs),				// 1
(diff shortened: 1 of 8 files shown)


The-Gig:
About the "no rocket jump if you have hook", does it also check that the aas file has been compiled with -grapplereach option? Otherwise it would be useless.

However, I'm not really sure the "no nocket jump if you have hook" is really useful: currently, you cannot really enable bot_grapple as it's unfinished and badly bugged. So telling bots to use the grapple in the place of rocket jump is not very useful, if they cannot use grapple in general.

Also, I don't know how the code currently is, but maybe the bot "thought" algorythm should consider something like:
1) I want to got up there (e.g. to take armor bonus), but it's too long to get there by walking.
2) Can I use grapple (bot_grapple, having it, aas supporting it)? If yes:
3a) If in aas there are grapplereach areas that allow me to get there from there, then check if there are enemies attacking me (using grapple requires time). If I'm not under attack, then use the grapple.
3b) If there are no grapplereach areas to get there, check if I can perform rocket jump (quantity of health, etc.) and check if there are weapon jump areas to get there. Then rocket jump.
Just a thought, I may be wrong.

NeonKnightOA:
Tell me two things:
a) what should be the ***exact*** purpose of bot_grapple
b) ***where*** does it need to be finished

Ambiguity is NEVER helpful when it comes to feedback.

The-Gig:
Well.. spectate bots with bot_grapple enabled in maps compiled with grapplereach option, such as oacmpdm10, and you may notice them trying to use it when they do not have it, shooting at higher walls with other weapons. And if they have it, sometimes they do shoot it at the same spot multiple times in a row, I'm not sure why they do abort the attempt and try again. As these are major problems, I don't know if there are other problems.

About how/when they are supposed to use it _exactly_, probably only some old id software employee knows... I can only guess. I guess as basic use they should use it to reach upper places if they consider it quicker than taking classic routes to go there.

The-Gig:
As secondary use, humans may use grapple  also horizontally, e.g. to pass to the other side of a gap (or even to the other side of the map, think about oacmpctf3)... but I have no idea if at id software they had also this use in mind for bots.

As third use, humans can use it to _remain attached_ to a wall after reaching it (note: remain attached to jumppad kills you quickly), even switching to a weapon and firing at people while staying attached (pro: surprise attack, con: you're an easy target once spotted). I don't know if they planned even this kind of use for bots, maybe it's too much!

The-Gig:
About grapple, I guess that just having bots using it for basic usage (using it to reach higher places) would be a great goal. I would suggest trying to fix those two annying bugs and then see what happens.

A thing I think would be **important** in case we fix bot_grapple, would be to _rename the cvar_ (bot_useGrapple maybe?), as mods inherit their initial configuration from baseoa, and if in baseoa bot_grapple would be set to 1, then mods would have it enabled... but the feature would still be broken there (except maybe a few mods with theoretically may have fixed the boot hook behavior by their own -I don't know if there are-).

NeonKnightOA:
Here's a test pk3, up to commit d0cc77c:
https://www.dropbox.com/s/seqsh3qwylgemdf/z_botai_test_oax.pk3?dl=0

Things that need to be tested:
- Domination (general behavior)
- Possession (general behavior)
- Rocket jumping behavior
- Grappling hook behavior
- Just in case there are problems, 1FCTF behavior

To do:
- The problem reported by Sub-Zero in Discord, regarding the bots not using MG in long distances. I **think** I've found the culprit, but I need to be sure.

NeonKnightOA:
Redoing the branch again.

The-Gig:
Just a note about "no rocketjump if you have grapple": from the first tests http://openarena.ws/board/index.php?topic=5458.msg55946#msg55946 with grapple it looks like the number of places where bots actually use grapple are quite limited. So I don't know if there how many chances there are that grapple really allows the bot to go to where it wanted to go via rocketjump.
Anyway, it looks like both features are quite rarely used by bots, so probably not a game changer.

PR #65, opened 2019-06-03
Refactorization of the botchat functions

NeonKnightOA:
So, a basic summary:
- Some checks done in every function that calls to specific chats were moved to more general functions. This should make the code cleaner.


NeonKnightOA:
Closing. I'll be working on a new PR where all of these chamges are included.

PR #66, opened 2019-06-03
Added a check (bot has the grapple?) for bot rocketjumping.

NeonKnightOA:


The-Gig:
Considering the reasons for using grapple instead of rocketjump to reach an higher place they are interested to go into, apart of course checking that the bot has got the grapple, bot_grapple is enabled and the .aas file has been compiled with -grapplereach option.... does it check it has actually got a way/route to reach the item it's interested with the grapple?

Maybe I explain better with an example. You have a powerful item placed on the top of a column. If .aas file says there is "weapon jump area" there, then a bot can rocketjump directly from the floor to the top of the column. But in case of grapple, you cannot grapple directly to the top of the column, if you are on the floor: you have to reach some other place first (e.g. the roof or a wall next to a platform near to the colums) and then go to the top of the column, by jumping or using grapple again (from an higher place)... if you are a bot, if the .aas file says there are grapple points there.

Uhm... thinking again about it, I can guess these checks are what the AI should evaluate to decide if it can reach a place using grapple *in general*, which may be a different function elsewhere and here invoked simply as "can I reach it with grapple? (yes/no)"...

NeonKnightOA:
Yes, this is a series of check done in order for a bot to decide if they should use the grapple. Actual grapple handling should be done somewhere else. In a separate pull request.

NeonKnightOA:
Closing. I'll be introducing these small changes in a new PR.

PR #67, opened 2019-06-03
Grappling hook handling has been moved to a separate function.

NeonKnightOA:


The-Gig:
One thing which still puzzles me is that it looks like if botfile CHARACTERISTIC_GRAPPLE_USER value isn't above 0,5, the bot will *never* use grapple... when I thought instead that was meant as a probability, so in theory in idsoftware's idea, even a bot with that value set to 0.1 may still have (rarely) used the grapple...

NeonKnightOA:
Closing. I'll be introducing these changes in a single PR in a new PR.

PR #68, opened 2019-06-03
Domination-specific fixes for botplay.

NeonKnightOA:
To-Do: Fix the ordering.

### DIFF
--- a/code/game/ai_cmd.c
+++ b/code/game/ai_cmd.c
@@ -142,6 +142,16 @@ void BotPrintTeamGoal(bot_state_t *bs) {
 			BotAI_Print(PRT_MESSAGE, "%s: I'm gonna take care of point B for %1.0f secs\n", netname, t);
 			break;
 		}
+		case LTG_DOMCAPTURE:
+		{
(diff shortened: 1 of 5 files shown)


sago007:
Crashes for me when launching a game on oa_shouse. The map does not have any domination points but the game should not have crashed. I notice that you divide by level.domination_points_count, however the game still crashed after I removed that, so there must be something else.

NeonKnightOA:
Weird, the 0.8.8 version of the map should have three points. I need to check this.

The-Gig:
_IIRC_ (!!!) the map was listed for domination in arenas file, but it didn't actually show any domination point...

sago007:
It does not really matter for this issue. The game should not break on a map as long as the map contains a spawn point.

NeonKnightOA:
It seems something that should be solved with a sanity check.
Can't check right now. But I guess that checking if the amount of domination points is below 1 should suffice to solve any problem regarding crashes.

NeonKnightOA:
I've added a sanity check which makes the bots to just roam in case of no points, as well as removed an unnecessary function.

NeonKnightOA:
Closing for the moment. I think it will be better to analyze how the AI code works and what can be done before any support can be introduced.

PR #70, opened 2019-06-03
[Testing required] Developer mode with two modes for coders and mappers.

NeonKnightOA:
There were a lot if #ifdef DEBUG in the code. So what about a mode where debug info could be displayed?

**[Here's a test pk3](https://cdn.discordapp.com/attachments/405518544382590987/675370157132087306/z_oax_v2.pk3) Updated February 7, 2020.**

We use `"developer 1"` to enter devmode for coders and `"developer 2"` for devmode for mappers. `"developer 3"` enters both devmode for coders as well as for mappers. Enabling devmode turns regular bot chat off (they would only input whatever information is needed for debugging).

### DIFF
--- a/code/game/ai_dmnet.c
+++ b/code/game/ai_dmnet.c
@@ -103,7 +103,7 @@ void BotRecordNodeSwitch(bot_state_t *bs, char *node, char *str, char *s) {
 	ClientName(bs->client, netname, sizeof(netname));
 	Com_sprintf(nodeswitch[numnodeswitches], 144, "%s at %2.1f entered %s: %s from %s\n", netname, FloatTime(), node, str, s);
 	// Developer mode outputs a message.
-	if (bot_developer.integer == 1) {
+	if (bot_developer.integer & DEVMODE_FOR_CODERS) {
 		BotAI_Print(PRT_MESSAGE, "%s", nodeswitch[numnodeswitches]);
 	}
 	numnodeswitches++;
@@ -160,7 +160,7 @@ int BotGoForAir(bot_state_t *bs, int tfl, bot_goal_t *ltg, float range) {
 	//if the bot needs air
 	if (bs->lastair_time < FloatTime() - 6) {
 		// Developer mode outputs a message.
-		if(bot_developer.integer == 1) {
+		if(bot_developer.integer & DEVMODE_FOR_CODERS) {
 			BotAI_Print(PRT_MESSAGE, "going for air\n");
 		}
 		//if we can find an air goal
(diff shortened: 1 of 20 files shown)


sago007:
Lines like
`if(!cg_developer.integer == 1)`
makes no sense as it is identical to:
`if(!cg_developer.integer)`.
A choice like Gig suggested with:
`if(cg_developer.integer&1)`
would be better. And the "1" should be a define.

The-Gig:
_Just a question_: are you sure to auto-silent bots while in developer mode? Cannot the dev just set bot_nochat if he doesn't want the chat? (Note: it's years I don't try bot_nochat, I do not even remember if that also applies to team orders....)

NeonKnightOA:
bot_nochat should be automatically activated when in developer mode, yeah. That's an extra check.

NeonKnightOA:
The check for bot_nochat is immediately after the check for developer. If it isn't in Developer mode, then players can set bot_nochat if they're bothered with the botchat.
So that's already tackled.

The-Gig:
What I meant was that maybe one dev might want to have both developer and both chat active at the same time, so maybe still having the chance to see bot chat while in developer mode sometimes might have been useful.
However probably it isn't so important, don't worry if you think it's not necessary. As I said, mine was just a question.
I don't know if "developer" enables some additional bot-chat-related warnings, so I don't know if it would be useful at all to enable it for bot chat debugging. I'm writing from phone, it isn't comfortable to check the code to see there are bot-chat-related developer messages. If there are not, then it's everything okay.

NeonKnightOA:
Debug messages for both mapper and coder modes are both enabled for developer mode.

code/game/ai_dmnet.c
```diff
@@ -2413,11 +2411,9 @@ int AINode_Battle_Retreat(bot_state_t *bs) {
 		AIEnter_Seek_LTG(bs, "battle retreat: enemy dead");
 		return qfalse;
 	}
-	//if there is another better enemy
-	if (BotFindEnemy(bs, bs->enemy)) {
-#ifdef DEBUG
+	//if there is another better enemy (only available in Developer mode)
+	if (BotFindEnemy(bs, bs->enemy) && (bot_developer.integer & DEVMODE_FOR_CODERS)) {
```
KuehnhammerTobias: BotFindEnemy isn't a developer function! That's why it wasn't inside #ifdef DEBUG!

NeonKnightOA: Oops... fixing! Thanks for reporting!

NeonKnightOA:
[Here's a test pk3](https://cdn.discordapp.com/attachments/405518544382590987/675370157132087306/z_oax_v2.pk3)

NeonKnightOA:
Doesn't seem to go anywhere. Closing.

PR #72, opened 2019-06-05
Replacement of the _IsAGametype functions for their defines

NeonKnightOA:
Exactly what it says on the title. With the define in place, we don't need those functions anymore.


NeonKnightOA:
Too early for this. Maybe some time later. Closing.

PR #73, opened 2019-06-05
Many changes for ai_team.c

NeonKnightOA:
* Simplified the code by sending redundant pieces of code to new function BotAI_GiveRoleToTheBot().
* Deleted BotDDOrders_Standard(). Its content is now inside of BotDDOrders. We shall see in the future.
* Ifs and Elses properly coded.

### DIFF
--- a/code/game/ai_team.c
+++ b/code/game/ai_team.c
@@ -427,9 +427,9 @@ void BotCTFOrders_BothFlagsNotAtBase(bot_state_t *bs) {
 		{
 			//tell the one not carrying the flag to attack the enemy base
 			if (teammates[0] != bs->flagcarrier) {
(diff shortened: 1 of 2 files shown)


KuehnhammerTobias:
@NeonKnightOA Hello! Sorry to contact you again but I think there is some problem with your code changes, for example you changed the 'BotCreateGroup' function so that now 'many' teammates (which isn't really true in this case) will accompany and the voice chat is 'follow flagcarrier'!
This doesn't make sense because 'BotCreateGroup' is used in Team Deathmatch game mode, how should a bot follow a flag carrier in TDM?
Sorry if I missed some changes to OA that this function is now used in gamemodes with a flag carrier, but it seems this is not the case.
I have to admit that I don't really understand what you want to fix with those code changes. Moreover simplification is not always a good choice and sometimes error prone, even more it is more about coding style/personal preference imo. Other people/developers prefer other coding style/format.

PR #74, opened 2019-09-24
ai_team.c: Code simplification by sending redundant pieces to new functions

NeonKnightOA:
This commit aims to make bot team formations easier to create by moving redundant pieces of code to a new function, as well as creating new bot orders.

For the future, we should create functions for Single Domination and more formations for Double Domination.

In general:
* Simplified the code by sending redundant pieces of code to new function BotAI_GiveRoleToTheBot().
* Deleted BotDDOrders_Standard(). Its content is now inside of BotDDOrders. We shall see in the future.
* Calls to the voicelines and chatlines were grouped by role. This should also make easier to add more roles in the future (such as those for Domination).


NeonKnightOA:
Closing. I'll be working instead on a unified branch where I'll be introducing small changes like this one.

PR #75, opened 2019-09-24
Replacement and deletion of redundant function TeamPlayIsOn() with the equivalent we already made.

NeonKnightOA:


NeonKnightOA:
Closing. I'll be integrating this and other changes on a separate PR.

PR #78, opened 2020-01-20
More cases to prevent bugs with gametypes and certain items.

NeonKnightOA:
Expansion of the previous pull-request to cover other gametypes and entities. This should fix most of the stuff.

### DIFF
--- a/code/game/bg_public.h
+++ b/code/game/bg_public.h
@@ -135,8 +135,8 @@ typedef enum {
 	
 } gametype_t;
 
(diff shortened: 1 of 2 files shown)


NeonKnightOA:
I was working in the wrong function all this time! Now every gametype should work as expected.
Also had to modify a bit bg_public.h, since it was taking Possession as a non-key-objective gametype.

NeonKnightOA:
Well, this commit should solve all bugs related to key objectives. In the end, I had to add new comparisons, since not all obelisk-based games use the neutral obelisk.

Once this is committed, I'd like to start working in the division between round-based gametypes and elimination-based gametypes, in order to fix a few things.

NeonKnightOA:
This doesn't seem to go anywhere, so I'm closing it.

PR #81, opened 2020-01-25
Better foundations for the AI.

NeonKnightOA:
This commit is composed of many small pull requests combined in a single file. These requests, and similar requests, are done with the idea of improving the AI, as well as giving it proper bot logic in all gametypes. However, for that to be a reality, first we need to lay some solid foundations. These commits aim to remove unnecesary and obsolete functions, group redundant pieces of code in new functions and add support and logic for the areas that require them the most.

The requests that made the initial commits are:
#75 (removal of redundant function TeamPlayIsOn(), since we have a proper validation for team games)
#65 (refactorization of ai_chat.c which allows better handling of the different chatlines)
#74 (refactorization of ai_team.c, which allows better handling of different bot formations in different situations)
#66 and #67 (handling of grappling hook usage and rocket jumping)

### DIFF
--- a/code/game/ai_dmq3.c
+++ b/code/game/ai_dmq3.c
@@ -1748,6 +1748,8 @@ void BotCheckItemPickup(bot_state_t *bs, int *oldinventory) {
 					if ((bs->ltgtype != LTG_GETFLAG) &&
 							(bs->ltgtype != LTG_ATTACKENEMYBASE) &&
 							(bs->ltgtype != LTG_HARVEST) &&
+							(bs->ltgtype != LTG_POINTA) &&
+							(bs->ltgtype != LTG_POINTB) &&
 							((!(GAMETYPE_USES_RED_AND_BLUE_FLAG(gametype) && !GAMETYPE_USES_WHITE_FLAG(gametype))) ||
 							((bs->redflagstatus == 0) &&
 							(bs->blueflagstatus == 0))) &&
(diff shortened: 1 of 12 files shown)


NeonKnightOA:
Undocumented in ai_vcmd.c are the actual changes:
- Replacement of _IsAGametype functions for their defines as well as added new checks.
- Initial DD/DOM support wherever needed.

NeonKnightOA:
Well, enough work for the day. For most of the code I couldn't really come with better solutions, however at least it compiles. Most of the support for Domination, Possession and Double DOM is already in place, as well as many other fixes. Sorry @sago007 , I couldn't come with anything better than this. 😓😅

Here's the test pk3, I'm very exhausted after working on this, so I couldn't test (tagging @The-Gig ): https://cdn.discordapp.com/attachments/405518544382590987/670818580841103370/oax.pk3
What needs to be tested? At least one match in DD, DOM and POS with 1, 2 and 3 bots, for a start.

For the future, I'd like to replace all the direct calls to gametypes (i.e. gametype == GT_DOUBLE_D) with rule calls (i.e. GAMETYPE_USES_OBELISKS(gametype)). That way, in the event a new gametype is introduced or a new change intended to affect many gametypes at the same time is introduced, it shouldn't be necessary to change much code.

NeonKnightOA:
Closing until all the issues are solved.

code/game/ai_dmq3.c
```diff
@@ -3567,7 +3597,9 @@ void BotAimAtEnemy(bot_state_t *bs) {
 	//
 	if (wi.number == WP_MACHINEGUN ||
 			wi.number == WP_SHOTGUN ||
+			wi.number == WP_NAILGUN ||
```
KuehnhammerTobias: Actually this was only meant for instant hit weapons, can you explain why you added the nailgun here?

PR #84, opened 2020-01-27
Grappling hook handling checks for bots

NeonKnightOA:


The-Gig:
Looking at the diff, I noticed a (yet to be completed) section about using grapple in case of falling into a deadly pit. Right under it there is the code about teleporting away if the bot has got personal teleporter. So I just wanted to notice that if it has bot both items, it should still prefer personal teleporter, which is immediately effective: I find it a bit unlikely to switch to grapple and attach somewhere in time, starting from hearing *falling1.wav (I think that's what triggers the action, right?), also considering that the sound often plays when not so far from the trigger_hurt. Maybe you already considered this, but I don't know enough about C to tell from reading the code.

*Maybe* a similar thinking may also apply to being in lava, which -unless you have environment suit- probably kills you faster than you may switch to grapple and attach aomewhere.

NeonKnightOA:
Oh, yeah, if bots have the tele they should use it first, indeed!
As for lava handling, I'll tackle that in another commit. Right now I have enough things to be done.

PR #85, opened 2020-01-29
Cleanup of the AI code

NeonKnightOA:
In this PR I'll be submitting fixes that involve removing annoyances such as single-line functions and ifdefs that prevent legit chunks of (necessary) code from running in the game. There won't be any refactor, any implementation of gametype support... all of that will have to wait because all of this is urgent and necessary!

As the title says, this will be limited to the ai_*.* files.

The last commit of this PR will involve the fusion of all headers into a single ai_local.h file.

Once all is said and done I'll remove the [WIP] tag.


NeonKnightOA:
This branch is set.

NeonKnightOA:
For this it would make more sense to send everything as individual commits rather than separate ones.

PR #87, opened 2020-01-31
Stripped filelists from the Windows batch files

NeonKnightOA:
Next: do the same but for the Linux scripts.


sago007:
I would rather just delete the "windows_script" folder.
The windows scripts files are for people that can't setup a proper environment. I do not want the rest of the build process soiled by it.

Back in the days getting a development environment setup was quite difficult but today with Windows Subsystem for Linux it is easy to compile the qvm files even on Windows.

Issue #86, opened 2020-01-30
[Feature] g_elimination_enableItems


NeonKnightOA:
Idea taken from [this post on the OA forums](http://openarena.ws/board/index.php?topic=5399.0). After getting used to the code, it shouldn't be that different, plus we have already an antecedent of a similar gamemode working, and it's [Unreal Tournament 4's (Team) Showdown mode](https://www.youtube.com/watch?v=cSxpCuBxEdc).

Implementation would be similar to Sago's suggestion: a single bitflag cvar.
**g_elimination_enableItems & 1 (ELIMINATION_ITEMS_ENABLEWEAPONS)** = enable
**g_elimination_enableItems & 2 (ELIMINATION_ITEMS_ENABLEHEALTHARMOR)** = but no health/armor
**g_elimination_enableItems & 4 (ELIMINATION_ITEMS_ENABLEPOWERUPS)** = but no power ups

The-Gig:
But in that case, what would a value of 5 do?

If it will be a bitflag variable, then maybe each value may *add* something instead, allowing maximum customization?
E.g.
1=add weapons/ammo boxes
2=add heath/armor bonuses
4=add powerups (also runes in case of g_runes 1?)

So to have everything one would set it to 7 (1+2+4)?

(On a side note, wondering if ui3 may allow to manage bit field settings in a way more comfortable than ""cycle between all possible combinations by repeatedly clicking").

NeonKnightOA:
Yep, my bad, I don't know what I was thinking when I wrote that.

The-Gig:
Just a small nitpicking... IIRC all elimination* tuning options do not have the initial "g_", probably to save some chars considering that "elimination" itself is quite a long word.
So, the cvar might be something like elimination_enableItems, elimination_keepItems, or something similar, without the g_.

Danil-Dm:
In my mod there is cvar elimination_items it includes items.
You can take ideas from my mod
https://github.com/Danil-Dm/openarena

NeonKnightOA:
I saw Ratmod implemented this very feature. I might have to take a look.

PR #91, opened 2020-02-03
[Testing required] elimination_keepItems

NeonKnightOA:
Here's the promised initial request for elimination_keepItems that appeared in #86.

[Here's also a test pk3.](https://cdn.discordapp.com/attachments/405518544382590987/673724175290204180/z_oax_v2.pk3)

In order to keep consistency across all of the Elimination modes, I had to create a new function that resumes all the checks for an Elimination-based ruleset.

What needs to be tested here?

* Item spawn in Elimination, CTF Elimination and LMS (in current master OAX there's a bug in LMS where items are spawned, this PR also solves that bug).
* Item spawn in Instagib.
* Item spawn in All Rockets.
* Item spawn in Elimination_AllGametypes.

Ideally, only key objectives should spawn and work in these modes. And in the case of Elimination, items other than key objectives should spawn only if **elimination_keepItems** is enabled.

**Once we tackle all of the bugs and quirks, then we can proceed to separate bunches of items, as mentioned #86.**

### DIFF
--- a/code/game/ai_main.c
+++ b/code/game/ai_main.c
@@ -188,8 +188,7 @@ int BotAI_GetEntityState( int entityNum, entityState_t *state ) {
 	memset( state, 0, sizeof(entityState_t) );
 	if (!ent->inuse) return qfalse;
 	if (!ent->r.linked) return qfalse;
-	if ( !(G_IsARoundBasedGametype(g_gametype.integer) ||g_instantgib.integer || g_rockets.integer || g_elimination_allgametypes.integer)
-	       && (ent->r.svFlags & SVF_NOCLIENT) ) {
+	if (!G_HasEliminationRules(g_gametype.integer) && ent->r.svFlags & SVF_NOCLIENT) {
 		return qfalse;
 	}
 	memcpy( state, &ent->s, sizeof(entityState_t) );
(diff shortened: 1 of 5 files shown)


The-Gig:
I've finally done the first tests with the option. They aren't all those requested, but I had limited time.

**First test: g_gametype 8; elimination_keepitems 1.** 
It seems to work fine, if you restart the map after enabling it.
If you enable it without restarting the map then, it doesn't work well: it looks like items are not visible, but you can still get them if you go over them (and it looks like they then become visible at their next respawn). A new round isn't enough to make them spawn, you need a new match.
Maybe you could search for a fix... or maybe you may just make the variable "latched", requiring a map restart to be effective.
Furthermore, if would be nice if it could print the server message to players when the value of the variable is changed, like other game-rules-changing variables.

**Second test: g_gametype 0; elimination_keepitems 1; g_elimination 1**
It seems to not work. Items do not spawn and you can't grab them.

About the other things you mentioned in the opening post... you mentioned instantgib and allrocket modes... do you mean that you planned the option to have effect also on those two modifiers, even if they are not "elimination"?

NeonKnightOA:
The checks for Elimination-like rules should also affect g_instantgib and g_allRockets. It makes no sense to have spawnable items in those modes. They should work the way they should do, and also receive the benefits of regular Elimination games (start with the intended weapons, etc).

The-Gig:
*Maybe* tomorrow I may do *a little bit* of testing, although my spare time is still very limited in the days (sorry guys)... do you have a new pk3 with the fixes for the problems found in the first tests?

NeonKnightOA:
Not yet. In fact I still need to update the branch with the latest changes.

NeonKnightOA:
Closing by now. There's an important commit which must be done first before keeping the work on this one.

PR #92, opened 2020-02-03
[Ready] Backports from ioQuake3 (gamecode)

NeonKnightOA:
ioquake/ioq3@c61417b8: Add current (custom) resolution to Q3 UI video mode list
ioquake/ioq3@9fcb2bb3: Override video mode list in Team Arena UI
ioquake/ioq3@f717a839: Fix lightning gun handling for corpses and single player podiums
ioquake/ioq3@dc0c3e7b: Fix team orders menu not listing clients with lower clientnums
ioquake/ioq3@a6e66700: Fix duplicate team join center print for bots and g_teamAutoJoin

### DIFF
--- a/code/cgame/cg_weapons.c
+++ b/code/cgame/cg_weapons.c
@@ -1760,7 +1760,7 @@ void CG_AddPlayerWeapon( refEntity_t *parent, playerState_t *ps, centity_t *cent
 	}
 
 	// make sure we aren't looking at cg.predictedPlayerEntity for LG
-	nonPredictedCent = &cg_entities[cent->currentState.clientNum];
+	nonPredictedCent = &cg_entities[cent->currentState.number];
 
 	// if the index of the nonPredictedCent is not the same as the clientNum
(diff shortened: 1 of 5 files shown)


The-Gig:
The description of the lightning gun thing resembles a glitch I noticed in 2010: http://openarena.ws/board/index.php?topic=3578.msg33231#msg33231

But in my tests back then it seemed that "firing ligthning from dead body" could happen *even before respawning* (idk if due to Elimination mode, which vanilla q3 does not have of course), and chaoticsoldier added that maybe that happened with gauntlet, too.

The-Gig:
I tried current nightly build code about that lighting glitch in Elimination mode. It looks like it's *partially* fixed now.
Before, you could still fire your lightning gun from your corpse even after the new "inactive warmup" had started. But before the new "inactive warmup" started, the lightning gun still fires from your dead body even with current gamecode.

NeonKnightOA:
Please, create a new issue with the bug and the steps to reproduce it.

PR #93, opened 2020-02-04
Fixed issues found with cppcheck

NeonKnightOA:
The issues are mentioned in each commit.

### DIFF
--- a/code/cgame/cg_weapons.c
+++ b/code/cgame/cg_weapons.c
@@ -1119,7 +1119,7 @@ void CG_RegisterItemVisuals( int itemNum )
 
 	item = &bg_itemlist[ itemNum ];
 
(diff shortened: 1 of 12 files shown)


sago007:
It all looks good to me but why
`memset( itemInfo, 0, sizeof( *itemInfo ) );`
instead of just
`memset( itemInfo, 0, sizeof( itemInfo ) );`?

NeonKnightOA:
Full explanation of the error was this:
```
[cgame/cg_weapons.c:1122] (warning) Size of pointer 'itemInfo' used instead of size of its data. This is likely to lead to a buffer overflow. You probably intend to write 'sizeof(*itemInfo)'. [pointerSize]
```

PR #88, opened 2020-02-01
[Testing required] Domination AI improvements

NeonKnightOA:
I took a look at the AI gamecode in order to see how can I improve bot support for OA's newer gamemodes. This specific PR focuses on Double Domination and Domination.

**[Test PK3](https://www.dropbox.com/s/mkgjyglgqyldi7q/z_oax-classicui-domsupport.pk3?dl=0). _Updated March 20, 2022._ Also requesting code review as I'm sure there are still some rough edges to polish.. - [Related forum thread](http://openarena.ws/board/index.php?topic=5458.0).**

**What needs to be tested?**

* Botmatches in Domination.
* How bots react to orders.
* If the bots do indeed go to pick the DOM points.
* Anything else that I'm missing.

**Additional notes**

- Currently the bots in Domination are set to pick a random point and hold onto it.
- As for their behavior, **the bots will pick a random point (if it falls inside of both established map and gametype limits -WIP-) and act upon it.** If the point is already taken, they can choose to hold onto that point, or go chase after another empty/enemy-controlled point. If all the points are taken, the bot will pick one at random and hold onto it. **The point of interest changes every time they're given an order or enter the game.**
* **Absence of control points will make the bots to roam through the map**. Also behavior accounts for the presence of a team leader, last goal, aggression and preference.

**To-Do**

* **Bots take a lot of time to reach the points and often get distracted in fights.** They should go straight to the CP.
* A feature I'd like to add, yet I haven't thought about how to put it in the code: **if an info_camp entity is linked to a DOM point or a flag, the bot should use it as a defense point**. Obviously the "angles" key should be pointing to the direction the bot should cover while defending.
* While checking ai_team.c, I figured out that bots have different team formations for aggressive and passive modes. I'd like to implement those as well in these modes.
* **Additional checks for gametype limits.**

PR #94, opened 2020-02-05
[Testing required] Enhanced AI checks for Grappling Hook handling

NeonKnightOA:
Well, what it says in the tin.

**[Here's a test pk3](https://www.dropbox.com/s/hu0fd56xzy06hp4/z_oax-classicui-grapplinghookforbots.pk3?dl=0). _(Updated March 20, 2022, 18:34 GMT-3)_ - [Related forum thread](http://openarena.ws/board/index.php?topic=5458.0).**

**Overview**

While in Q3A (and up to OA 0.8.1) the Grappling Hook was left as just a dummied out weapon, post-0.8.1 it has been gradually rebuilt as a weapon, filling the blanks where needed until the weapon achieved a somewhat acceptable "functional" status in 0.8.8. However, there are many problems with it regarding AI handling:

> The grapple AI logic for bots has many bugs: it's unclear if they try to pick-up the grapple or not... if they fire the grapple and miss the edge of the target wall, they try again without correcting the aim... if they have to go to a place where they would have used grapple but don't have it yet, they start pointlessly firing at that place with standard weapons.
> -- [OA Wiki](https://openarena.fandom.com/wiki/Bugs) - Also related OA forum posts: [1](http://openarena.ws/board/index.php?topic=1908.msg47648#msg47648) - [2](http://openarena.ws/board/index.php?topic=4679.msg47628#msg47628)

I can make the checks required for the bots to stop trying to get the Grapple when they don't have it, but there are stuff which I don't really feel confident to do. Any kind of help in the form of code edits or snippets is welcome.

**What needs to be tested?**

* AI behavior when using the Grappling Hook

**Additional notes:**

In the current implementation (of this PK3), **bots will use the Grappling Hook if (in order):** the general grappling setting for them is ENABLED, they HAVE the GH in their inventory and their personal bot settings ALLOW for its usage.

### DIFF
--- a/code/game/ai_dmnet.c
+++ b/code/game/ai_dmnet.c
@@ -1557,7 +1557,8 @@ int AINode_Seek_ActivateEntity(bot_state_t *bs) {
 	}
 	//
 	bs->tfl = TFL_DEFAULT;
-	if (bot_grapple.integer) bs->tfl |= TFL_GRAPPLEHOOK;
+	// if they have the Grappling Hook
+	if (BotCanAndWantsToUseTheGrapple(bs)) bs->tfl |= TFL_GRAPPLEHOOK;
 	// if in lava or slime the bot should be able to get out
 	if (BotInLavaOrSlime(bs)) bs->tfl |= TFL_LAVA|TFL_SLIME;
 	// map specific code
@@ -1770,7 +1771,8 @@ int AINode_Seek_NBG(bot_state_t *bs) {
 	}
 	//
 	bs->tfl = TFL_DEFAULT;
-	if (bot_grapple.integer) bs->tfl |= TFL_GRAPPLEHOOK;
+	// if they have the Grappling Hook
+	if (BotCanAndWantsToUseTheGrapple(bs)) bs->tfl |= TFL_GRAPPLEHOOK;
 	//if in lava or slime the bot should be able to get out
 	if (BotInLavaOrSlime(bs)) bs->tfl |= TFL_LAVA|TFL_SLIME;
 	//
@@ -1913,7 +1915,8 @@ int AINode_Seek_LTG(bot_state_t *bs)
 	}
 	//
 	bs->tfl = TFL_DEFAULT;
-	if (bot_grapple.integer) bs->tfl |= TFL_GRAPPLEHOOK;
+	// if they have the Grappling Hook
+	if (BotCanAndWantsToUseTheGrapple(bs)) bs->tfl |= TFL_GRAPPLEHOOK;
 	//if in lava or slime the bot should be able to get out
 	if (BotInLavaOrSlime(bs)) bs->tfl |= TFL_LAVA|TFL_SLIME;
 	//
@@ -2191,7 +2194,8 @@ int AINode_Battle_Fight(bot_state_t *bs) {
 	BotBattleUseItems(bs);
 	//
 	bs->tfl = TFL_DEFAULT;
-	if (bot_grapple.integer) bs->tfl |= TFL_GRAPPLEHOOK;
+	// if they have the Grappling Hook
+	if (BotCanAndWantsToUseTheGrapple(bs)) bs->tfl |= TFL_GRAPPLEHOOK;
 	//if in lava or slime the bot should be able to get out
 	if (BotInLavaOrSlime(bs)) bs->tfl |= TFL_LAVA|TFL_SLIME;
 	//
@@ -2284,7 +2288,8 @@ int AINode_Battle_Chase(bot_state_t *bs)
 	}
 	//
 	bs->tfl = TFL_DEFAULT;
-	if (bot_grapple.integer) bs->tfl |= TFL_GRAPPLEHOOK;
+	// if they have the Grappling Hook
+	if (BotCanAndWantsToUseTheGrapple(bs)) bs->tfl |= TFL_GRAPPLEHOOK;
 	//if in lava or slime the bot should be able to get out
 	if (BotInLavaOrSlime(bs)) bs->tfl |= TFL_LAVA|TFL_SLIME;
 	//
@@ -2421,7 +2426,7 @@ int AINode_Battle_Retreat(bot_state_t *bs) {
 	}
 	//
 	bs->tfl = TFL_DEFAULT;
-	if (bot_grapple.integer) bs->tfl |= TFL_GRAPPLEHOOK;
+	if (BotCanAndWantsToUseTheGrapple(bs)) bs->tfl |= TFL_GRAPPLEHOOK;
 	//if in lava or slime the bot should be able to get out
 	if (BotInLavaOrSlime(bs)) bs->tfl |= TFL_LAVA|TFL_SLIME;
 	//map specific code
@@ -2600,7 +2605,8 @@ int AINode_Battle_NBG(bot_state_t *bs) {
 	}
 	//
 	bs->tfl = TFL_DEFAULT;
-	if (bot_grapple.integer) bs->tfl |= TFL_GRAPPLEHOOK;
+	// if they have the Grappling Hook
+	if (BotCanAndWantsToUseTheGrapple(bs)) bs->tfl |= TFL_GRAPPLEHOOK;
 	//if in lava or slime the bot should be able to get out
 	if (BotInLavaOrSlime(bs)) bs->tfl |= TFL_LAVA|TFL_SLIME;
 	//
@@ -2694,4 +2700,3 @@ int AINode_Battle_NBG(bot_state_t *bs) {
 	//
 	return qtrue;
 }
-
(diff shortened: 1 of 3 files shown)


The-Gig:
Excuse me NK, could you please re-read your comment above? There are a couple of places where a bold phrase seems to contradict a non-bold phrase in the same paragraph.

In the meanwhile, a random thought: PT and GH are not fully interchangeable... for emergency escape PT would usually be better (e.g. standing still while waiting for the grapple to reach may be your end if someone is attacking you), but PT would cost you your objective item you are carring... and there are some situations where only grapple is adapt (as grapple brings you where you wish, while PT is random destination).
I think bots would probably mostly use grapple as a means to quicky reach upper places as a shortcut to reach some item they are interested (like they do rocket jump) than as an escape manouvre...

Just to say that having both items does not autmatically mean one should *always* take priority... id depends from context. Probably I'm Captain Obvious here, but just to be sure I said this.

NeonKnightOA:
I've reworded most of the post. I agree some parts were confuse.

Gig, a little reminder: these checks are used in order to decide **whether a bot should use the Grapple/PT or not**. The actual grapple/PT usage is handled somewhere else.

The-Gig:
About your "to-do" above: how to plan detect that they are falling into a deadly pit? Checking if their fall will end up in a trigger_hurt which has got a damage value bigger than X? Maybe the *falling1 check may still remain as an OR?
And *I don't know* if there will be need to do checks to prevent bots from using grapple while falling in a deadly pit: I don't know how the -grapplereach bspc option works exactly, but I can *guess* (it's just a theory!!!) it would create a series of "A->B" associations (specific starting areas tied to specific destination areas), and bots would use grapple only in those specific places... so *probably* bots would not use grapple to reach places they would get trapped and maybe they would not even use it when they are mid-air at all... Of course, considering bot_grapple was broken, it's hard to tell!

The-Gig:
Uh?
*"bots will use gh if.... the general grappling setting for them is off"*
Off? Did you mean on?

*"They will always use GH in lava or slime"*
I would think that if they can get out by just walking/running/jumping in a short time, that may be better than using the gh, which is quite time consuming. Idk if a such check would be easy or not.

NeonKnightOA:
Damn me! I was looking directly at the code and got everything wrong! Fixing!

The-Gig:
BTW, also RJ isn't always interchangeable with GH. Of course, *if the aas files says that the bot can use both means to reach the place it wants*, then the grapple is better as it doesn't hurt.

KuehnhammerTobias:
Just a small note, it is good habit to set travelflags before using them. Here (inside 'BotCanAndWantsToUseThePersonalTeleporter') you check if the bot is in lava, afterwards you set the travelflag for if he is in lava. Keep in mind bot thinking are server frames (CPU cycles), so try to write efficient code, which is very often not the style we humans prefer :)
Though, this doesn't make your actual changes work, it just saves you some CPU power.
Your code doesn't work because of the fact The-Gig already said: The items aren't interchangeable. So this is a 'AI-logic' thing... The Grappling Hook areas are stored in the AAS file and only the Grappling Hook will bots to go for those areas (hence the travelflag for the Grapple) etc. The same is true for rocket jumps.

NeonKnightOA:
Well, I've removed the Teleporter handling checks (incidentally I've stumbled upon with the area that, well, handles them).

[Here's another test pk3](https://cdn.discordapp.com/attachments/405518544382590987/679415622047891473/z_oax_v2.pk3).

NeonKnightOA:
Excised all references to the Personal Teleporter.

The-Gig:
The link to download the test pk3 from three years ago seems to be broken.

NeonKnightOA:
Oh, yeah, I was supposed to re-upload it again. I'll do it now.

KuehnhammerTobias:
Once more, since a bot still checks if he has a teleporter he will NOT use the grapple if he has a teleporter.
Hmm, I think you somehow mean that a 'return' statement tells the bot to do the next thing 'in order'. This is NOT the case. As soon as you 'return' from a function bots will not check the next 'option'. In other words, as soon as something inside a function tells the bot that he can't or should not use the grapple (or any other following function) he will not look for another item. Keep in mind a return statement terminates the execution of a function.

NeonKnightOA:
DAMN ME! I accidentally force-pushed the branch from another PC I was working from!
Reverting to just Grapple.

NeonKnightOA:
There! @KuehnhammerTobias, how should it look now?

KuehnhammerTobias:
What I wrote above is true for Rocket Jumping as well. A check that returns to false will stop the bot from RJ.
Again:
The bot can/should reach an area ONLY with the RJ -> he checks if he has the RL -> oh, I have the RL in my pocket, cool -> now he checks if he has the grapple too -> oh, sadly I am also wearing a grapple -> so unfortunately I can't RL :(
It is important to understand the code! Do you know what I mean?

KuehnhammerTobias:
Btw, always when you play around with the bot characteristics, be aware that you will eventually ruin custom made bots and their AI.
For example, default characteristics like 'grapple user' or 'self-preservation' are already used very specific, and custom bot makers are used to the default behaviour.
As soon as you say that bots with a value of < 0.5 will stop the bots from using the grapple, will change/ruin the current bot behaviour.

NeonKnightOA:
Oh, yeah, BotCanAndWantsToRocketJump. It was what inspired me to do the GH checks in order to stop bots from grappling constantly when they have the Grapple.

> The grapple AI logic for bots has many bugs: it's unclear if they try to pick-up the grapple or not... if they fire the grapple and miss the edge of the target wall, they try again without correcting the aim... if they have to go to a place where they would have used grapple but don't have it yet, they start pointlessy firing at that place with standard weapons.

NeonKnightOA:
I've just removed the GH check from RocketJump.

BTW, @KuehnhammerTobias , can I ask you what are your thoughts about #88? It's the one PR I'm most concerned about and the one I want to polish and finish as soon as I can.

The-Gig:
You know I'm an ignorant person... but I'm still puzzled by the fact that a bot file setting value in <0.5 resulting in "*never* use grapple" instead of "*rarely* use grapple"... do the other "CHARACTERISTIC_" in bot personality files work the same?

NeonKnightOA:
That characteristic IICR can take any value between 0 and 1 (such as 0.75). So if a bot has that characteristic over 0,5 they will use the grapple.

NeonKnightOA:
Added a little overview to the PR, also updated the test pk3.

I'm still not confident enough to actually do anything outside of basic checks. So any kind of help in the form of code snippets or where to look at is welcome.

KuehnhammerTobias:
There aren't many GPL v2 projects with AI code you can have a look at. afaik. If I remember correctly RWG mod and Brainworks mod both are GPL v2, maybe have a look at them, here is the Brainworks repositiory: https://github.com/TimeDoctor/quake3-brainworks

NeonKnightOA:
@KuehnhammerTobias AFAIK, somebody asked the creator of Brainworks to release their mod under GPLv2 for an OA-related project some time ago (I guess aftershock made use of the mod).

@The-Gig In many areas of the code some of the bot characteristics are checked against a constant value for the sake of reducing calculations, as many of these could just be solved with a simple "yes/no" instead of a complex calculation.

NeonKnightOA:
Also CHARACTERISTIC_GRAPPLE_USER, while being defined, isn't used anywhere in the code at least as far as "master" is concerned. And I'm sure this is also true of even VQ3 as well.

KuehnhammerTobias:
Neither The-Gig nor I said to NOT us CHARACTERISTIC_GRAPPLE_USER at all. The problem is like The-Gig already said above, it's not clear if it is a good idea to completely disable using Grapple by a value of < 0.5, that's all.
Btw, just to avoid a rumor, using characteristics isn't a complex calculation, I don't know where you get this information from.
The bot characteristics just define some individual values concerning bot behaviour.
You have five of them bounded and unbounded integers and floats, and string.
Using trap_Characteristic_Integer acts more or less than 1/0 (or like you said yes/no), so you don't need to be concerned about that. Even a float characteristic of 0.5 works like 100 * 0.5 = 50 (not really complex I would say.

NeonKnightOA:
So this means this fragment of the RocketJumping function is also wrong (I've based the GH function on this after all):
```
/* ai_dmq3.c - BotCanAndWantsToUseRocketJump() */
	rocketjumper = trap_Characteristic_BFloat(bs->character, CHARACTERISTIC_WEAPONJUMPING, 0, 1);
	if (rocketjumper < 0.5) return qfalse;
```
Well, I guess I shall try something else.

NeonKnightOA:
Testing PK3 updated.

NeonKnightOA:
Guess I'm just going to redo this from scratch.

PR #96, opened 2020-02-05
[WIP] Removal of _UsesKeyObjectives and _UsesControlPoints. Also added new comparison _UsesTheNeutralObelisk.

NeonKnightOA:
In addition, unlocked some game logic which used to need MP but it doesn't make sense now.


NeonKnightOA:
It just got too big with a lot of changes. Closing and opening separate PRs for each change.

PR #98, opened 2020-02-06
Replacement of direct array length calls by GT_MAX_GAME_TYPE constant call.

NeonKnightOA:


sago007:
I find ARRAY_LEN a lot safer. I can instantly see that the next line does not cause undefined behaviour.

NeonKnightOA:
So basically, the desired is the inverse: reverting GT_MAX_GAME_TYPE to the array length comparison?

sago007:
> So basically, the desired is the inverse: reverting GT_MAX_GAME_TYPE to the array length comparison?
Only in cases where the check is used to validate array bonds. I would not use it anywhere else.

NeonKnightOA:
Indeed.

PR #102, opened 2020-02-07
Unification of AI headers into a single ai_local.h header file

NeonKnightOA:


NeonKnightOA:
On second thought, this will only make things harder. Closing for the moment.

PR #103, opened 2020-02-07
Unification of key objective carrying AI behavior into a single function

NeonKnightOA:
This allows us to have an unified behavior when a bot carries a key objective (flags, skulls) and goes towards its objective (or roams, in case of Possession).


NeonKnightOA:
Not worth messing with.

PR #105, opened 2020-02-10
Fix: "Next round" Elimination countdown is also shown in Double Domination.

NeonKnightOA:
Back in Unreal Tournament 2004, after any team scored a point, the game would tell the players with a countdown when the next round was going to start. "New round in... 5... 4... 3... 2... 1...".

This small change does that for our version of DD as well.


sago007:
Tested it. Did feel better.

PR #107, opened 2020-02-10
[WIP] Gametype comparisons by fraglimit, by capturelimit, by scorelimit and by harvestlimit

NeonKnightOA:
So, several gametypes require comparisons based on their score limit.
Currently grouped limits are:
- Fraglimit (FFA, 1on1, TDM, LMS)
- Harvestlimit (Harvester)
- Scorelimit (Domination/Possession)
- Capturelimit (everything else)

Why separate score limits for Harvester, DOM and POS? Because capturelimits tend to be too short (and fraglimits too long, sometimes) for Harvester, which usually requires limits near 30 for a lasting game. Also capturelimits and even fraglimits tend to be too short for DOM and POS, which uses score limits that go beyond 100 points.

The newer gametypes also are votable, like fraglimit. I was wondering why capturelimit wasn't votable.

Defaults across every module: Frag 15, Cap 5, Harvest 30 and Score 100.


The-Gig:
It's many years I suggest to make capturelimit votable.
That's the only change I would suggest to do about this. Once you can vote for capturelimit, and the vote menu also foresees some high values like those used in DOM and POS, you don't need a ton of different cvars.
Too many "xlimit" cvars involve more voting options, *a longer serverinfo string* (and when it reaches a certain lenght, strange things start happening... and at a certain lenght it stops working at all IIRC. And the game does not warn you about reaching those limits.), additional server cvars for setting max and min votelimit for each of them.
Also, *IIRC* there are already some ui_ cvars which allow you to set a different "default" value for each gametype when starting a match from the GUI.

In this case, I would opt for "keep it simple": I suggest to just add the ability to callvote for capturelimit, add its max/min cvars, and add some high values to the vote menu. Simple as.
Note: years ago a guy (who later left the forum deleting all of his posts) did make a such try, and **IIRC** he noticed there was a character lenght limit for voting, so he had to name the voting option "ctflimit" or "caplimit" (I would prefer the latter) because "callvote capturelimit" was too long...

The-Gig:
@sago007 Any thoughts?

NeonKnightOA:
For the moment I've removed the harvestLimit and assigned harvester's score limit to Fraglimit (as I've said before, making it a capturelimit would make games extremely short). Might consider to do that as well for Domination and Possession, but I'm not sure.

NeonKnightOA:
Closing for the moment.

PR #108, opened 2020-02-15
Several fixes and improvements for code reading

NeonKnightOA:


code/game/g_local.h
```diff
@@ -1431,4 +1431,5 @@ qboolean G_IsATeamGametype(int check);	/* Whether the gametype is team-based or
 qboolean G_UsesTeamFlags(int check);	/* Whether the gametype uses the red and blue flags. */
 qboolean G_UsesTheWhiteFlag(int check);	/* Whether the gametype uses the neutral flag. */
 qboolean G_IsARoundBasedGametype(int check);	/* Whether the gametype uses the neutral flag. */
+qboolean G_HasEliminationRules(int check);	/* Whether the game uses Elimination rules (all weapons + no pickups). */
```
sago007: Drop the parameter and just check for g_gametype.integer inside the function. Having an argument signals to the reader that the function does not read cvars directly and that is false.

NeonKnightOA: I'll remove the HER commit from this branch (it's also related to #91) and make it into a separate branch so we can polish it from there. What about the rest of the commits, are they OK?

NeonKnightOA: Done. I'll open a separate PR for that one.

NeonKnightOA:
On second thought, I should better prepare a proper code reading fix before anything else.

code/game/g_local.h
NeonKnightOA: There, the function has been moved to #111.

PR #111, opened 2020-02-17
Fix: New function G_HasEliminationRules() contains gametypes/settings which use Elimination-like rules

NeonKnightOA:


sago007:
I believe that this change decreases readability. G_HasEliminationRules is way to complicated. If I need to look at the source of a helper function then the helper function does not give any value.

PR #112, opened 2020-02-18
BotBattleUseItems replaced with four different functions which handle each of the holdables

NeonKnightOA:
Why? Because the old function was just a function calling other functions, and the remains could be sent to two different functions. It was a waste of function stack.

### DIFF
--- a/code/game/ai_dmnet.c
+++ b/code/game/ai_dmnet.c
@@ -2147,8 +2147,13 @@ int AINode_Battle_Fight(bot_state_t *bs) {
 			return qfalse;
 		}
 	}
+
 	//use holdable items
-	BotBattleUseItems(bs);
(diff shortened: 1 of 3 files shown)


KuehnhammerTobias:
Just a note: you don't fix anything here! You change bot behaviour, because you now introduce their self preservation! I highly recommend to NOT call your PRs a 'fix', the same is true for #119. Calling a PR a fix, suggests that something is broken. Just a suggestion.
As far as this PR is concerned, please note that a bot with a self preservation < 0.5 will NOT use their item(s) anymore, this makes them acting very silly, becaue why should the bot not use the medkit for example, if he is carrying this item. I don't see what will you fix with this PR?

NeonKnightOA:
I thought I've removed the self preservation part. It was an idea but didn't work as expected.
The overall idea is to remove a function which just calls other functions. And the remaining part of the code could be made into two separate functions as well.

PR #113, opened 2020-02-19
Fix: g_spSkill could take floating point values. This small change should work.

NeonKnightOA:


sago007:
It looks to me that the value end up in "bot_character_t *BotLoadCharacterFromFile(char *charfile, int skill)" in the engine and will be converted to an int, so even if the value is a float all the way down, I do not see that this would help.

The-Gig:
It passed some time since this... anyway, for what I remember, in Q3A/OA 0.8.8 you set g_spkill a skill level with a fractional part (something like /g_spskill 3.5; idk if map_restart is required) and then if you check the info of a bot _later added by bot_minplayers_ (like /dumpuser sarge), its skill is reported including the fractional part. I don't know if a skill of 3.1 does actually differ from 3.9, up to this moment I just assumed so.
Anyway, in this comment https://github.com/OpenArena/gamecode/commit/1542c068f0b4fd2c4099847dbfa3aa0d6d5cb9b4#commitcomment-37370196 I noticed that in OAX after https://github.com/OpenArena/gamecode/commit/1542c068f0b4fd2c4099847dbfa3aa0d6d5cb9b4 this commit, dumpuser on bots added by bot_minplayers do not show the fractional part anymore (unlike bots added manually, which still do).

The-Gig:
By the way, time ago I did try to figure out whether, even with previous gamecode, fractional parts in bots skill levels actually changed something... _I wasn't able to reach a definitive answer_ (e.g. by not being able to detect whether something changed in bots aiming when changing e.g. from 3.1 to 3.9), but at least looking at the output damage they caused or how much damage they could sustain (using g_debugdamage), it _looked_ to me that the damage didn't change.

NeonKnightOA:
I need to close this temporarily.

PR #119, opened 2020-02-23
Replacement of unnecessary function BotChatTime() with a define.

NeonKnightOA:
It returns a fixed value, so it makes no sense to have a function that just returns a fixed value which isn't even calculated.


sago007:
Looks good to me

PR #122, opened 2020-02-23
Fix: Classic menu now accounts for "Hide Private"

NeonKnightOA:


sago007:
Looks good to me.
Except the default value should be true (1).

NeonKnightOA:
Done.

PR #125, opened 2020-02-25
Changed the commenting style for those annoying ending brackets.

NeonKnightOA:
These brackets in their old form would prevent the folding functions of many editors such as Notepad++ or SciTE from properly working.


NeonKnightOA:
Closing for the moment.

PR #126, opened 2020-02-25
New cvar: g_criticalHit. A random shot may have the chance to inflict 3x damage.

NeonKnightOA:
Not intended to be a serious cvar, OFC.
The game informs when somebody did this upper left (haven't properly checked cgame though).

### DIFF
--- a/code/game/g_combat.c
+++ b/code/game/g_combat.c
@@ -1265,6 +1265,21 @@ void G_Damage( gentity_t *targ, gentity_t *inflictor, gentity_t *attacker,
 		damage *= g_damageModifier.value;
 	}
 
+	if(g_criticalHit.integer) {
+		float check1,check2,check3,check4,check5;
+
+		check1 = rand() % 5;
+		check2 = rand() % 6;
+		check3 = rand() % 7;
+		check4 = rand() % 8;
+		check5 = rand() % 9;
+
+		if (check1 > 2 && check2 > 3 && check3 > 4 && check4 > 5 && check5 > 6) {
+			trap_SendServerCommand(-1, va("print \"%s did a CRITICAL HIT!\n\"", attacker->client->pers.netname));
+			damage *= 2;
+		}
+	}
+
 	if ( damage < 1 ) {
 		damage = 1;
 	}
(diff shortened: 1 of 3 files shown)


The-Gig:
What's the probability of the critical hit? And is the probability based upon the number of *shots* or the number of *hits*?

If one wants to make it "seriously" (for how "serious" a such feature may be in a game like this), maybe the probability should change depending from the weapon you are using, as different weapons have different fire rates. Which might be a fitting probability, like one every two full loads of shots (e.g. one every 20 shotgun shots, one every 20 grenades shot, one every 100 plasma gun shots)? And what about gauntlet, and TA weapons (which, unlike other weapons, on picking they give fewer shots than their ammo boxes)?

You made me think about "what about a sort of *doom damage mode*"? Do you remember that in Doom each shot could randomly make a damage varying from 1x to 8x multipliyer (e.g. if a weapon had a base value of 3, then each one could randomly cause either 3, 6, 9, 12, 15, 18, 21 or 24 damage). Of course our interval might be smaller. This probably is not very adapt for a competitive "arena" fps, but the same may be said for critical hits. Maybe both of them are more adapt for a mod than baseoa... but in case you like them both, maybe both might be enabled using a single bitfield cvar, like g_randomDamage, g_crazyDamage or something similar.

General thought: considering the room for new cvars is limited, maybe throwing in anything which comes to our minds isn't the best thing ever...

NeonKnightOA:
Ok, I'm closing this.

Gig, a piece of advice: feedback is welcome, except when it boils down to "why do you care?" or "why do we need this?"

The-Gig:
The reason for doing something should be one of first things to think about.
You can know better than me about cost/benefit in terms of resource usage, performances, the risk of causing new bugs, etc.

I'm NOT saying "don't do any changes"... let's try to think about what we need...

The-Gig:
And anyway I didn't dismiss your idea... I did even propose a way to make it coherent between different weapons...

If you want to post a pk3, I'll test it.

The-Gig:
Anyway, you said the criticalhit does 3x damage, but in the code I see "damage *= 2"... doesn't that mean 2x damange?

PR #128, opened 2020-02-28
Fix: cg_weaponBarStyle with value outside of range now displays default weapon bar.

NeonKnightOA:
Previously it would hide the weapon bar completely.

### DIFF
--- a/code/cgame/cg_weapons.c
+++ b/code/cgame/cg_weapons.c
@@ -2300,8 +2300,7 @@ void CG_DrawWeaponSelect( void )
 	}
 
 	switch(cg_weaponBarStyle.integer) {
-	case 0:
-		CG_DrawWeaponBar0(count,bits);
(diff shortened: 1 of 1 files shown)


The-Gig:
I think the difference between 5 and 7 is that with 7 the background slowly changes from blue to red as your ammo goes down.

Also, I don't know if to use -1 or 8 for "no weapon bar"... What do you think existing users would expect more?

The-Gig:
See also: https://github.com/OpenArena/gamecode/issues/143

sago007:
I like that the weapon bars are now described by comments.

NeonKnightOA:
I hope the comments I wrote were accurate, at least.

The-Gig:
Uhm... Maybe you forgot to update the comments of those that at first look seem the same,  but with the background that changes slowly? See the second post of this discussion.

PR #130, opened 2020-02-29
New fields for .info files

NeonKnightOA:


The-Gig:
Excuse me, what do you mean with map "title"?
The map longname maybe? But that is alredy in .arena file, isn't it?
Do you mean this way one can type it either in one file or the other one?

NeonKnightOA:
Yep, the longname. It was weird to me that all other fields are there and this one wasn't. Though since .arena and arenas.txt are still a necessary thing for compatibility...

The-Gig:
Thank you!

But with your last sentence, do you mean that maybe it would be better to make .info files only contain data which is not already included into .arena files?

NeonKnightOA:
.info files take precedence over .arena and arenas.txt. It could help make arenas.txt smaller. arenas.txt could contain only the relevant, necessary data, especially with many maps. (map filename and supported gametypes).

The-Gig:
Good point. Very well!
That would have some drawback with mods I guess... howerver it's still an option we may consider in case we will lack space in arenas.txt.

NeonKnightOA:
I've figured out this can be extended. Some other fields may be useful.

PR #131, opened 2020-03-07
Unified default scorelimit and timelimit values / Deleted unused cvars in MPUI

NeonKnightOA:


The-Gig:
Excuse me, is there some relationship between this stuff and the various ui_ffa_timelimit, ui_tournament_fraglimit etc. etc. Cvars?

PR #133, opened 2020-03-09
Fix: flags won't appear anymore in Harvester/Overload. Also flag bases re-added to CTF/eCTF/1FCTF/Possession.

NeonKnightOA:


NeonKnightOA:
Splitting this one too.

PR #147, opened 2022-03-21
Gametype aliases for .info files

NeonKnightOA:
Split of #130, as it deserves its own PR.

I'd like to extend this to eventually .arena support and "gametype"/"!gametype" key support.


sago007:
There is no reason to allow alternatives. Fewer is better. The only reason to allow multiple choices is to preserve backwards compatibility after a name fix.

PR #148, opened 2022-03-22
widescreen fix

EddieBrrrock:
Deletion of a patch for widescreen resolution that cause some problems


EddieBrrrock:
Before :
Capture_decran_2022-02-17_a_16 28 25

After :
Capture_decran_2022-02-17_a_16 28 09

PR #150, opened 2022-03-27
New fields for .info files

NeonKnightOA:
Redoing this commit. `.info` files will be important for the future of OA, so we should take care of them.

* `title` is the title of the map (i.e. `oa_bases3.info` would have `title "Some Bases"`)
* `mpBotCount` is the amount of random bots that should be summoned (i.e. `mpBotCount 3`)
* `redBots` and `blueBots` are the bots that should be summoned for the Red and Blue teams in team games (i.e. `redBots "merman sarge"`)
* `redBotCount` and `blueBotCount` are the number of random bots that should be summoned and added to the Red and Blue team in team games (i.e. `redBotCount 3`)
* `recommendedTeamSize`, much like `recommendedPlayers`, is the amount of ideal members on each team for the map (i.e. `recommendedTeamSize 3`)

**To-Do:**

Implement functions that return particular data. These functions will look first in the `.info` files, then in `.arena` files, then in `arenas.txt`:
```
char* GetMapName(const char* filename);
char* GetMapAuthor(const char* filename);
char* GetMapDescription(const char* filename);
int GetMapBotCount(const char* filename); // amount of random bots to summon
char* GetMapBotNames(const char* filename); // list of names of particular bots to summon
int GetMapRedBotCount(const char* filename); // amount of random bots to summon and add to the Red team
char* GetMapRedBotNames(const char* filename); // list of names of particular bots to summon and add to the Red team
int GetMapBlueBotCount(const char* filename); // amount of random bots to summon and add to the Blue team
char* GetMapBlueBotNames(const char* filename); // list of names of particular bots to summon and add to the Blue team
int GetMapFragLimit(const char* filename);
int GetMapTimeLimit(const char* filename);
int GetMapCaptureLimit(const char* filename);
int GetMapMinTeamSize(const char* filename);
int GetMapMaxTeamSize(const char* filename);
int GetMapRecommendedTeamSize(const char* filename);
int GetMapMinPlayers(const char* filename);
int GetMapMaxPlayers(const char* filename);
int GetMapRecommendedPlayers(const char* filename);
```
For the record, `.arena` files and `arenas.txt` are looked for in these files: `game/g_bot.c` (function `G_LoadArenas`), `q3_ui/ui_gameinfo.c` (function `UI_LoadArenas`) and `ui/ui_gameinfo.c` (function `UI_LoadArenas`).

### DIFF
--- a/code/game/bg_misc.c
+++ b/code/game/bg_misc.c
@@ -1889,11 +1889,29 @@ void MapInfoGet(const char* mapname, int gametype, mapinfo_result_t *result) {
 		if (Q_strequal(keyBuffer, "author")) {
 			if (mayRead) Q_strncpyz(result->author, token, sizeof (result->author));
 		}
+		if (Q_strequal(keyBuffer, "title")) {
+			if (mayRead) Q_strncpyz(result->title, token, sizeof (result->title));
+		}
 		if (Q_strequal(keyBuffer, "description")) {
 			if (mayRead) Q_strncpyz(result->description, token, sizeof (result->description));
 		}
 		if (Q_strequal(keyBuffer, "mpBots")) {
 			if (mayRead) Q_strncpyz(result->mpBots, token, sizeof (result->mpBots));
+		}
+		if (Q_strequal(keyBuffer, "mpBotCount")) {
+			if (mayRead) result->mpBotCount = atoi(token);
+		}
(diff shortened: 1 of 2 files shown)


sago007:
The new fields looks fine to me.
You might want to hold back on implementing the helper functions until they are needed.

PR #153, opened 2022-04-03
Cleanup of unused cvars in MPUI/UI3

NeonKnightOA:
These cvars weren't used anywhere in the code of MPUI/UI3 (they originated in q3_ui). Deleting them for the moment.

### DIFF
--- a/code/ui/ui_local.h
+++ b/code/ui/ui_local.h
@@ -32,39 +32,6 @@ Foundation, Inc., 51 Franklin St, Fifth Floor, Boston, MA  02110-1301  USA
 
 // global display context
 
-extern vmCvar_t ui_ffa_fraglimit;
-extern vmCvar_t ui_ffa_timelimit;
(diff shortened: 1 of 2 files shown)


The-Gig:
Wondering how this may affect testing in classic oax (iirc oax_m -ui3- is still to much "alpha" to really be useable) and if it might affect existing mods (also considering the case of a blank config)...

NeonKnightOA:
Classic OAX isn't affected at all as the cvars are still there (check the q3_ui folder for that). This only concerns the MP UI.

NeonKnightOA:
Exactly. The four relevant folders for the gamecode are code/cgame (clientside) code/game (serverside) code/q3_ui (classic but hardcoded Q3-like UI) and code/ui (the backend side of the TA/MP/UI3 script-based, more flexible UI, the frontend would be UI3 which is located on its own repo).

NeonKnightOA:
Closing for the moment.

PR #154, opened 2022-04-03
Checks for the AI use of the Grappling Hook (Simplification of #94)

NeonKnightOA:
It prevents the bots from using the grapple if they aren't allowed to do so, if they don't have the weapon, and if their AI settings would prevent them from doing so.

This is a cleaner implementation of #94 that cuts the nonsense and goes straight to the point.


NeonKnightOA:
Fuck! I've just noticed something that's wrong with this PR!

Fixing ASAP!

PR #155, opened 2022-04-03
Fix: CTF Flags won't be drawn anymore in Harvester/Overload

NeonKnightOA:
This PR only affects the team flags. The neutral flag/skull receptacle works well.

Here's how it looks in vanilla 0.8.8:
shot0002

Here's how it looks in latest OAX:
shot0001

And here's how it looks with this fix:
shot0004

### DIFF
--- a/code/game/g_items.c
+++ b/code/game/g_items.c
@@ -1006,6 +1006,11 @@ void G_SpawnItem (gentity_t *ent, gitem_t *item)
 		ent->s.eFlags |= EF_NODRAW; //Don't draw the flag models/persistant powerups
 	}
 
+	if( (g_gametype.integer == GT_HARVESTER || g_gametype.integer == GT_OBELISK) &&
+			(strequals(ent->classname, "team_CTF_redflag") || strequals(ent->classname, "team_CTF_blueflag") ) ) {
+		ent->s.eFlags |= EF_NODRAW; // Don't draw the team flags in Harvester/Overload
+	}
+
 	if( !G_UsesTheWhiteFlag(g_gametype.integer) && strequals(ent->classname, "team_CTF_neutralflag")) {
 		ent->s.eFlags |= EF_NODRAW; // Don't draw the flag in CTF_elimination
 	}


NeonKnightOA:
Added images.

NeonKnightOA:
Well, color me amazed. This is how it looks in Team Arena:

Neutral objectives


Flag bases


Now I'm not sure this should be done at all...

NeonKnightOA:
For the sake of it, a pic of mpteam1 from Team Arena, for comparison.


Yep, looks like this requires manual map fixes rather than code fixes. Closing.

The-Gig:
I think we should make the red and blue flag bases (red/blue "small" obelisks) models with some red and blue shades instead of just black (that would be an _OA assets_ fix)...

In the meanwhile, maybe keeping the flags for identifying red and blue bases may look nice in Harvester, but in Overload I fear it could be a mess (the flag compenetrating with the "big" obelisk), do you have a screenshot if that in OA?

NeonKnightOA:
[oac_czest2](https://www.moddb.com/games/openarena/addons/oac-czest2-a4v2-the-bridge-of-despair-remix), the map of the screenshots, has all the key items without their flags specified.

May be useful for testing reasons.

The-Gig:
I guess the fact that Team Arena doesn't automatically hide the flags in some mode does not necessarily imply that we should do the same...

The flags compenetrating with the Harvester obelisks look quite ugly in Team Arena (see your first screenshot here <https://github.com/OpenArena/gamecode/pull/155#issuecomment-1474815815>) and in OA too (see these images)...
shot0257
shot0256

I'm not sure whether it's better to let the task to the mappers or not....

NeonKnightOA:
Apparently it's meant for the mappers to do such things.

My problem is that I fail to see how a circling flag can be used as a prop in a map.

Then again, we aren't supposed to modify Q3 behavior. And that, unfortunately, includes the bugs.

The-Gig:
> Apparently it's meant for the mappers to do such things.

Okay...
> My problem is that I fail to see how a circling flag can be used as a prop in a map.

Same for me. As I said before, maybe it may help identifying the skull receptacles that in OA are black and quite small, but apart from that, I don't know...
> Then again, we aren't supposed to modify Q3/TA behavior. And that, unfortunately, includes the bugs.

While _I really agree with that as a general guideline_, I think we may _evaluate_ some exceptions for what's not core gameplay, especially in case of bugs and things which do not really change the gameplay. As an example, Sago changed some scoring rules so now your player score cannot go below 0 in some gamemodes and when there are only two players one committing suicide gives the point to the other one instead of subtracting one from him: they do somehow deviate from q3, but it a way meant to make it less frustrating... same applies to "award pushing" being enabled by default, I guess.

NeonKnightOA:
Reopened after consulting @sago007

NeonKnightOA:
I need to close this temporarily.

PR #156, opened 2022-04-03
Regression Fix: Obelisk bases are now drawn outside Harvester/Overload.

NeonKnightOA:
Useful for CTF maps that use them as flag bases. Plus, [the official Q3TA editing manual](https://icculus.org/gtkradiant/documentation/Team_Arena_Mapping_Help/pages/ta_game_types.html) recommends doing so, so there's no point in not drawing them outside of Harv/Over.

Using [this map](https://www.moddb.com/games/openarena/addons/oac-hydro3-a2) for testing.

Here's how it looks in vanilla, mod-less 0.8.8:
shot0000

Here's how it looks in current OAX:
shot0001

And here, how it looks with the fixed version:
shot0000

### DIFF
--- a/code/game/g_team.c
+++ b/code/game/g_team.c
@@ -2297,7 +2297,7 @@ void SP_team_redobelisk( gentity_t *ent )
 {
 	gentity_t *obelisk;
 
-	if (g_gametype.integer != GT_HARVESTER && g_gametype.integer != GT_OBELISK) {
+	if (!G_UsesTeamFlags(g_gametype.integer) && g_gametype.integer != GT_HARVESTER && g_gametype.integer != GT_OBELISK) {
 		G_FreeEntity(ent);
 		return;
 	}
@@ -2323,7 +2323,7 @@ void SP_team_blueobelisk( gentity_t *ent )
 {
 	gentity_t *obelisk;
 
-	if (g_gametype.integer != GT_HARVESTER && g_gametype.integer != GT_OBELISK) {
+	if (!G_UsesTeamFlags(g_gametype.integer) && g_gametype.integer != GT_HARVESTER && g_gametype.integer != GT_OBELISK) {
 		G_FreeEntity(ent);
 		return;
 	}
@@ -2347,7 +2347,7 @@ void SP_team_blueobelisk( gentity_t *ent )
 */
 void SP_team_neutralobelisk( gentity_t *ent )
 {
-	if (g_gametype.integer != GT_HARVESTER) {
+	if (!G_UsesTheWhiteFlag(g_gametype.integer) && g_gametype.integer != GT_HARVESTER) {
 		G_FreeEntity(ent);
 		return;
 	}


The-Gig:
Are you extending the support to modes where those bases would make sense or to everything? Just as example, I see no purpose to show neutral obelisk outside from oneflag and harvester...

NeonKnightOA:
> Are you extending the support to modes where those bases would make sense or to everything? Just as example, I see no purpose to show neutral obelisk outside from oneflag and harvester...

From the official Q3TA manual which indeed (as I guess you or sago pointed out) are used as a flag base.

https://icculus.org/gtkradiant/documentation/Team_Arena_Mapping_Help/pages/ta_game_types.html

The-Gig:
Yes, it says that obelisks will act as flag bases. But it also says "do not mark any gametype", and I think that probably takes in account the fact that the game filters them for you. For the fame stops doing that filters, mappers should start doing it, which would be an issue for old maps.

So, maybe Obelisks should only appear in the gametypes that, years ago, you pointed out in this table? <https://antifandom.com/OpenArena/wiki/Mapping_manual/Additional_gametype_support#Per-gametype_entity_support>

The-Gig:
Adding a screenshot of the table I mentioned above, for quicker reference.
Screenshot_20230310-233543

NeonKnightOA:
That table needs to be updated with Possession, heh.

Yeah, more or less I remember why I did that table in the first place.

The-Gig:
So, to sum up, I think it should be:
- Red/blue obelisks in ctf, oneflag, harvester, obelisk, ctfelimination, dd, dom (note: according to 088 documenation the gametype key value for ctfelimination is just "elimination", but probabky that's another part of code)
- Neutral obelisk in harvester, obelisk, pos.

In the next days I'm curious to check what it actually does in 088/current nightly builds gamecode.

NeonKnightOA:
I need to check ioq3 or the original q3a code to see what exactly I'm doing wrong.

Or old 0.8.8 with one of the current maps that don't have the key items marked with a gametype flag.

NeonKnightOA:
Yep. It's a regression.

I'm using [this map](https://www.moddb.com/games/openarena/addons/oac-hydro3-a2) for testing.

Here's how it looks in vanilla, mod-less 0.8.8:
shot0000

Here's how it looks in current OAX:
shot0001

So yeah, it's a regression.

The-Gig:
A regression, now I understand why this confused me!
I just tried to find out which older commit introduced the regression (mostly for curiosity, but also might be useful to see if it intrudced similar problems on other items), but it's harder than I thought... Especially from phone!

NeonKnightOA:
I'm fairly sure it was one I made. [Here are all the changes made to this file up to now](https://github.com/OpenArena/gamecode/commits/master/code/game/g_team.c).

My apologies!

NeonKnightOA:
Regression confirmed. This is how it looks like in Team Arena:


Also a pic of mpteam1, a default map from Team Arena:

The-Gig:
> My apologies!

Don't worry! :-)

PR #157, opened 2022-04-14
Widescreen fix

Bishop-333:
Deletion of a patch for widescreen resolution that cause some problems


EddieBrrrock:
Before :
159443489-4e4332f7-114b-449c-8e11-dde0864a3737
After :
159443535-d5cd9444-ece3-46b9-ac82-8168f26283cb

PR #161, opened 2022-05-15
Yourteamcapture.wav is now only played once when it's a personal capture instead of two before

EddieBrrrock:


The-Gig:
I found an old 0.8.x bug report which said "`flagcap.wav (you are on red team, and red team scores) plays half speed and half length on SDL, plays twice on SDL and OpenAL. Should be mono, not stereo?`"
(I guess they were two different bugs. The bit about mono/stereo thing may be related to engine instead, see also https://github.com/OpenArena/engine/issues/27)

PR #170, opened 2023-04-12
Domination AI Improvements

NeonKnightOA:
Added Botfile headers for compilation reasons and because they're integral for the PR to work.

Resume of the PR:
* In Developer mode, bots now mention that they're controlling a point.
* New field for bots: currentPoint, contains the actual DOM point the bot is caring about. Gets updated every time BotSetDominationPoint() is called, and BotGetDominationPoint() returns it. If BotSetDominationPoint() gets an invalid value, it instead picks up a random control point.
* When a bot is told to attack or defend a point, it assigns it a random point so the bot can go and hold it. Both commands assign a new control point for the bot to take.
* Team formations for bots (ai_team.c): All bots are given the order to roam.
* Support for Missionpack/UI3 voice commands.
* GUI orders for Classic UI.


NeonKnightOA:
There are some roughs around the edges, yeah. I wasn't too comfortable with attack and defend doing the same thing (rerolling the CP).

The-Gig:
What's the reason for bots reporting the name of the doination point on developer mode only? As long as it's a team-based chat, it would make sense to tell it also in normal matches, isn't it?
Maybe the reason is that it would be too often?

NeonKnightOA:
More for tracking purposes. If they target the same point too often, for example. And which point are they targeting.

NeonKnightOA:
Looks like this branch needs a lot of work before it can be committed. Looks clean but there's some kink to work out.

PR #171, opened 2023-04-25
Replacement of g_rockets with g_weaponArena and g_weaponArenaWeapon

NeonKnightOA:
This PR basically replaces and extends `g_rockets` in order to work with all weapons rather than only rockets as `g_weaponArena`. It's comprised of two cvars:

* `g_weaponArena`: 0 disables it, 1 enables it, 2 enables it and adds the Grappling Hook to the inventory.
* `g_weaponArenaWeapon`: specifies the weapon that should be added to the inventory. If the Grappling Hook is selected, then it returns the default value (Gauntlet).

The accepted table of values for `g_weaponArenaWeapon` are as follows:

| Weapon | Strings |
|:---|:---:|
| Machinegun | `mg`, `machinegun`, `machine gun`, `2` |
| Shotgun | `sg`, `shotgun`, `shot gun`, `3` |
| Grenade Launcher | `gl`, `grenade`, `grenadelauncher`, `grenade launcher`, `4` |
| Rocket Launcher | `rl`, `rocket`, `rocketlauncher`, `rocket launcher`, `5` |
| Lightning Gun | `lg`, `lightning`, `lightninggun`, `lightning gun`, `6` |
| Railgun | `rg`, `railgun`, `rail gun`, `7` |
| Plasma Gun | `pg`, `plasma`, `plasmagun`, `plasma gun`, `8` |
| BFG | `bfg`, `big fairie gun`, `big freakin gun`, `9` |
| Nailgun | `ng`, `nailgun`, `nail gun`, `11` |
| Prox Launcher | `pl`, `prox`, `mines`, `proxlauncher`, `prox launcher`, `12` |
| Chaingun | `cg`, `chaingun`, `chain gun`, `13` |

Any other value not in this list for `g_weaponArenaWeapon` spawns the Gauntlet. `g_weaponArena 2` spawns the Grappling Hook alongside the selected weapon.

While currently enabling the option via menu is quite limited (I haven't figured out how to add the weapon option in classic menu), the cvars are fully operative.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,6 +20,7 @@
   * `g_harvesterFromBodies`: In Harvester matches, skulls now spawn from dead bodies (a la [UT3!Greed](https://antifandom.com/unreal/wiki/Greed)) rather than a skull receptacle in the middle of the arena. Allows Harvester matches to take place in maps that don't feature a skull receptacle. **Default: 0.**
   * `g_ddCaptureTime` and `g_ddRespawnDelay`: New cvars for Double Domination that control the amount of holding time to score and the waiting time before a new round starts. **Default for both: 10.**
   * `bot_developer`: Bitflag that enables bot-based feedback such as action report (`&1`), regular chat disabling (`&2`) and obstacle feedback (`&4`). **Default: 0.**
+  * `g_weaponArena` and `g_weaponArenaWeapon`: two cvars that replace and extend `g_rockets` in order to be able to use `g_rockets` with every weapon other than rockets.
 * Shuffle has been reworked by implementing the solution from Aftershock.
 * New commands:
   * `weapbest`: Selects the best weapon.
@@ -36,6 +37,9 @@
 * Tons of other bug fixes.
 
 ### Extended version
+* New cvars that replace `g_rockets`: `g_weaponArena` and `g_weaponArenaWeapon`.
+  * `g_weaponArena` enables the "Weapon Arena" mode, which is the old g_rockets, but with any other weapon. `g_weaponArena 2` adds the Grappling Hook to the inventory.
+  * `g_weaponArenaWeapon` controls, via specific strings, which weapon it's spawned with its ammo boxes. The weapon cannot be picked up via menu, only (for now) via console. The full string list can be locate [here](https://github.com/OpenArena/gamecode/pull/171).
 * New cvar: bot_developer, for bot-based debugging.
 * Missionpack/UI3 backend refactors.
 * "You Have Been Mined" message outright displays the counter instead of delaying it.
--- a/code/cgame/cg_servercmds.c
+++ b/code/cgame/cg_servercmds.c
@@ -362,7 +362,7 @@ void CG_ParseServerinfo( void ) {
 	cgs.timelimit = atoi( Info_ValueForKey( info, "timelimit" ) );
 	cgs.maxclients = atoi( Info_ValueForKey( info, "sv_maxclients" ) );
 	cgs.roundtime = atoi( Info_ValueForKey( info, "elimination_roundtime" ) );
-	cgs.nopickup = atoi( Info_ValueForKey( info, "g_rockets" ) ) + atoi( Info_ValueForKey( info, "g_instantgib" ) ) + atoi( Info_ValueForKey( info, "g_elimination" ) );
+	cgs.nopickup = atoi( Info_ValueForKey( info, "g_weaponArena" ) ) + atoi( Info_ValueForKey( info, "g_instantgib" ) ) + atoi( Info_ValueForKey( info, "g_elimination" ) );
 	cgs.lms_mode = atoi( Info_ValueForKey( info, "g_lms_mode" ) );
 	cgs.altExcellent = atoi( Info_ValueForKey( info, "g_altExcellent" ) );
 	mapname = Info_ValueForKey( info, "mapname" );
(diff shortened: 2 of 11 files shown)


The-Gig:
Wait a minute, different values for g_weaponArena for with and without grapple does not sound optimal for me, as oax gamecode has already got g_grapple (which _replaces_ elimination_grapple).
_I suggest to just let g_grapple do its job also in weaponArena mode_ and you're already ok, no need to make things redundant. Otherwise, you end up with a mess like we have elimination_selfdamage which intersects with both g_friendlyfire and dmflags 8 and dmflags 1024. (See <https://antifandom.com/OpenArena/wiki/Elimination#CVARs> and <https://antifandom.com/OpenArena/wiki/Special_game_options#Dmflags> for some infos)
I suggest to try avoiding redundant/conflicting variables, at least for new stuff.

The-Gig:
@NeonKnightOA did you read that comment above, where I suggested to just let g_weaponArena be a simpler 0/1 and let g_grapple handle grappling hook?
Any thoughts?

The-Gig:
Wait, one more thing that just crossed my mind: what about consolidating g_weaponArena and g_weaponArenaWeapon in just a single cvar, instead of two? If it's emtpy or 0 (or an unrecognized value), the feature is disabled; if it's set to any recognized value, then that weapon is the one used. Some cvar memory space saved.
What to you think?

NeonKnightOA:
> @NeonKnightOA did you read that comment above, where I suggested to just let g_weaponArena be a simpler 0/1 and let g_grapple handle grappling hook? Any thoughts?

I read it. I've decided to continue with this since g_weaponArena only affects g_weaponArena and not the other modes.

> Wait, one more thing that just crossed my mind: what about consolidating g_weaponArena and g_weaponArenaWeapon in just a single cvar, instead of two? If it's emtpy or 0 (or an unrecognized value), the feature is disabled; if it's set to any recognized value, then that weapon is the one used. Some cvar memory space saved. What to you think?

I thought about it, and it was my original wish, but it would make a ''very'' long bitflag (and it would defeat the purpose of "one weapon arena" since that would lead to combinations of weapons and _that_ would have been quite complex to implement). For now, this is the way it'll be implemented.

The-Gig:
Uh? Why should it be a bitflag in the first place? :-/

NeonKnightOA:
You have:
* 1 - mode enabled
* 2 - adds GH
* 4 - MG
* 8 - SG
* 16 - GL
* 32 - RL
* 64 - LG
* 128 - RG
* 256 - PG
* 512 - BFG
* 1024 - NG
* 2048 - PL
* 4096 - CG

The maximum number would be 8191. For the purposes of the cvar, only 13 of those values matter.

The-Gig:
I'm a little lost, I'm sorry.
Do you mean that "g_weaponArenaWeapon" currently converts the value you type `rl, rocket, rocketlauncher, rocket launcher, 5` into `32` (10000 bin)? When?
I haven't checked the code, I assumed g_weaponArenaWeapon was a text field and there were some checks at round startup like _"if g_weaponArenaWeapon = (rl OR rocket OR rocketlauncher OR rocket launcher OR 5) then give rocketlauncher and remove everything else"_...

Uhm... maybe you mean that _for using a single cvar instead of two_, you would have made it a bitfield like the post above this one? But why? Couldn't the single cvar be a string (text) field?
Like:
* gauntlet, 1 - Only gauntlet
* mg, machinegun, machine gun, 2 - Only machinegun
* sg, shotgun, shot gun, 3 - Only shotgun
* rl, rocket, rocketlauncher, rocket launcher, 5 - You guessed it
* [...]
* cg, chaingun, chain gun, 13 - you guessed it
* None of the above (blank, 0, anything else) - Mode disabled

And grappling hook (technically weapon 10) managed via g_grapple.
This is what would make most sense to me. Maybe there are technical reason why this would be harder to code?

(On a side note, in order to simplify the checks, maybe just supporting the weapon number in theory could be already enough, and the cvar being an integer, not a string nor a bitfield).

NeonKnightOA:
A single cvar isn't an option in this case because one value (weaponArena) requires a number and the other (weaponArenaWeapon) strings, and both data types (integer and string) have different operations.

The-Gig:
To understand, why does g_weaponArena _need_ to be a number, and cannot be a string?

However, just accepting a _weapon number_ as the input, instead of a series of different text values for each weapon, could do the trick, I guess. (1=gauntlet, 2=machinegun ... 13=chaingun; 0, 10 or any other number=disabled).

The-Gig:
FatSakura said this on Discord:
> I did add g_weaponarena with one cvar a year ago in my mod idk if its pertinent or it can help :
> first commit : https://github.com/Bishop-333/OmegA-mod/commit/58f2d5004ad34f5c3602372e00c13fa1ca6d2f7a
> second : https://github.com/Bishop-333/OmegA-mod/commit/63a156a089fe77b0ded11176e8f0cb602d1d0024
> third : https://github.com/Bishop-333/OmegA-mod/commit/f36358b9fa47be69d350931bb6f93985bd5f8243#diff-420e8197c2e7866b8d10959599aa228fa0d6e53ded4975f44c3d98550fcfdcba

The-Gig:
The change of g_weaponarenaweapon from string to integer NK did here
https://github.com/OpenArena/gamecode/commit/3a11287220ffa764fa0b35bb0dff4407ccbe3ae6
Might be first step towards an optimization fo single cvar?

NeonKnightOA:
Not a chance. The reason is the same `g_doWarmup` and `g_warmup` exist as separate cvars.

PR #177, opened 2024-03-12
[WIP] New gametype checkers for scorelimits

NeonKnightOA:
We have gametype checkers handled by key game objective. These checkers handle support for scoring.


NeonKnightOA:
I'll better implement this later.

PR #192, opened 2024-03-13
Removal of unnecessary function CG_DrawStatusBarFlag, as it just called another function.

NeonKnightOA:


The-Gig:
Before you remove a function, you are 100% sure it isn't invoked anywhere anymore (gamecode, engine, classic ui, ui3), right?

NeonKnightOA:
It's just a function that just called another function. Cluttered the function stack unnecessarily. Such functions are safe to remove.

The-Gig:
I get it's like function A which just calls function B... Basically useless, but before removing function A, all calls to function A anywhere else should be updated to invoke function B directly, isn't it?

NeonKnightOA:
Exactly.

The-Gig:
Okay, well then. :)

But of course, please pay attention whether a function might potentially be invoked by other modules, in order to do not break compatibility with third party engines or mods. You know, if something apparently useless was there is the first place, maybe it was for compatibility reasons, who knows...

I'm not really sure about how such stuff works, but I have the feeling there could be a few mods which only change the UI while still relying on base gamecode. _I might be wrong_. I think FlagRaiders q3 mod may be a such case.

You know, this whole "underhood refactoring project" scares me a little...

NeonKnightOA:
Not a risk. OA doesn't have modular modding like the Unreal engine games, where you can just extend other classes and create smaller packages that can be loaded with the game itself as a result. Here you have to compile your own .vm/.so files and therefore you need access to the full gamecode source code.

For example, in the UE games you can just extend the Shock Rifle class and create a one-hit kill Shock Rifle and create a mutator to load it. All without modifying the game itself. In Q3/OA you have to create your own .vm/.so even for the smallest things since Q3/OA isn't that easily moddable. That's why a lot of mods have to be loaded in their own folder.

The-Gig:
Yes, I remember the Mutators in UT99. Pretty cool, you could mix "only shock rifles" with "big heads" and "bullet time"...

I know quake3/oa isn't as much flexible, but considering they exist things like almost-serverside-only mods with _optional_ revamped ui (if the client has got the pk3 he gets mod-custom HUD, if the client hasn't got it, he can still play on the server using the classic HUD) and I don't know how such stuff works exactly... I just ask (hoping to do not annoy to much, I like to learn). As far as I can tell, gamcode-side there should be a client part, a server part and an ui part...

Trying to look into the pk3 files of a bunch of mods...
I see in "Alternatefire" mod there are, under \vm, "cgame.qvm, qagame.qvm and ui.qvm"... but also in "Flagraiders2" there are the same three files, as the majority of mods have. In "unlagged" there is only cgame.qvm and in "rvr" there are only cgame.qvm and qagame.qvm.

NeonKnightOA:
Exactly. All these three files control the game logic. And if you place them in baseq3/baseoa, you're effectively replacing Q3/OA with that mod instead of complementing it.

The-Gig:
Okay. But my concern wasn't about placing a mod qvm in baseoa, I've always known that's a no-no.
What I was worrying about is if a mod can provide only one of two or them instead of all three of them, then removing or excessively altering (in baseoa) functions, variables or whatever potentially used cross-module might crash some mods? In case a certain function can only be invoked within its own QVM, then I guess it should be fine. Otherwise, it might be a risk, I suppose (and I guess server/client game logic somehow shares some data with the UI, otherwise how could the HUD show your inventory, just to make an example?).

NeonKnightOA:
_"In case a certain function can only be invoked within its own QVM, then I guess it should be fine."_

This is always the case with game mods. The only universal variables are found in the engine part.

The HUD is managed in cgame, which _is_ game logic. The engine only handles the rendering.

The-Gig:
Very well then. Thank you!

PR #199, opened 2024-03-15
Widescreen fix

Bishop-333:
Make the leilei widescreen fix cvar dependant to avoid code deleting and fix hud position :

**BEFORE :** 
Capture d’écran 2024-03-15 à 22 46 51
Capture d’écran 2024-03-15 à 22 59 35

**AFTER :** 
Capture d’écran 2024-03-15 à 22 49 35
Capture d’écran 2024-03-15 à 22 59 20

It caused some element to draw over others like the cg_drawteamhud drawing over the scores


Bishop-333:
PS : fps change is not related and cg_leiWidescreen was unused so setting it to 0 will have no other impact

NeonKnightOA:
Noted and committed.

leilei-:
fix the other elements instead.

Bishop-333:
I didnt mean your code deleting improvement but avoid the deletion of your code, but ok ill try get around and find a solution

PR #200, opened 2024-03-15
Openal sound restart

Bishop-333:
OpenAL setting change require a snd_restart so make it automatically as done when changing sound quality


NeonKnightOA:
Committed.

The-Gig:
Is this something already done elsewhere or is it an exception?
IIRC many cvars which require a snd_restart or vid_restart require you to do it manually (although it's possible that some sound cvars do it without me noticing, a vid_restart is more easily noticeable \*)

IIrc, an automatic vid/snd restart is more common when changing an option _from the GUI_ instead. I don't know whether this pull request is about changing the cvar directly or through the GUI.


\* Iirc, a snd_restart doesn't show the map load screen, but I might recall wrong and I can't start OpenArena at the moment.

The-Gig:
The commit seems to have made changes in q3_ui folder... So it's applied on changes from the gui? Then it's ok I guess. But what about the OA3 new ui? Does it have a sound settings menu yet?

Bishop-333:
@The-Gig snd_restart is something actually noticeable it shows the map loads to. Changing the sound quality actually does already this automatic snd_restart and on OpenAL settings it dont even shows if you have disabled it or not after clicking (button is still lightned up and its writted "on") before you do a snd_restart thats why i added this to avoid confusion.
For the new UI i know that atleast there is a marker/cross that shows when you disable it after clicking

@NeonKnightOA Why reverting it ?

The-Gig:
I just tried the Sound menu in OA 0.8.8.
Changing "sound quality" option _immediately_ causes a snd_restart (player experiences a reload screen and returns to gameplay).

Changing graphic options which require a vid_restart do not immediately cause a restart: they make an "apply" button appear. The "apply" button disappears if you change the option back to its former value.
The changes are effective only if and when you play the "apply" button, and are discarded if you leave the page without pressing it (settings which do not require a restart are immediately effective instead).

Thus, I think the "good way" to do this would be making _both_ "sound quality" and "OpenAL" options work the same as the "graphic" options menu: settings which require a restart should make an "apply" button appear. Am I wrong?

Bishop-333:
@The-Gig Can do that 👍 But maybe for the mouseover more like : OpenAL mode require a compatible engine 
Because except quake3e, OA, ioquake3 and omega engine have it already built in no ?

The-Gig:
Iirc, OA 0.8.8 doesn't come with Windows OpenAL library, I think.
And quake3e removed OpenAL support completely, IIRC...

NeonKnightOA:
Indeed, it doesn't come. You need to copy OpenAL32.dll from Windows/System32 into OA.

The-Gig:
Maybe I didn't copy the dll from Windows into oa folder, I think I did just install it in the system (using the old package from Creative Labs, probably). But that was many years ago, I'm not sure.

Anyway, what about "OpenAL may require manual install of its library. Some engines may not support it."?
I'm not sure whether we should consider the existence of third party engines or not.

PR #208, opened 2024-03-16
elimination_selfdamage is now elimination_damage, is a bitflag, and can be set in Classic Menu.

NeonKnightOA:


The-Gig:
Wouldn't it be better to plan/discuss changes a litte (here on on Discord) before doing them? :confused:

PR #307, opened 2024-03-27
"New round" message is now drawn for Overload.

NeonKnightOA:


The-Gig:
Notr: this wan't right, as Overload doesn't have Rounds. Reverted, Iirc.

PR #308, opened 2024-03-27
Ready effect in railgun glass will match his trail color

Bishop-333:
Blue team :

Color1 = green, color2 = magenta


Bishop-333:
You can test with this : 
[z_zrailgun-colors.pk3.zip](https://github.com/OpenArena/gamecode/files/14773071/z_zrailgun-colors.pk3.zip)

NeonKnightOA:
I'm committing this. The feature will be available in tomorrow's release.

The-Gig:
I finally tested this (using [2024-04-01_6f52109.zip](https://files.poulsander.com/~poul19/public_files/oa/dev088/gamecode_nightly/openarena_gamecode_2024-04-01_6f52109.zip)). Nice!
Two drawbacks I should point out, though:
- Every time you change your railgun colors, you have to relearn which one means "reloading" and which one means "ready".
- If you set both color1 and color2 to the same color, the only difference between "reloading" and "ready" is that the latter is a bit brighter.

PR #329, opened 2024-04-03
Add dmflags menu

Bishop-333:


NeonKnightOA:
Excellent! But before I commit this, may I request the menu to be summoned from Skirmish/Create Server rather than Setup?

The-Gig:
NK, maybe you might provide him a screenshot of the point where you would want the link, within the Create server page?

Furthermore, IIRC you were planning to move some settings to an Advanced page. Do you think it's better Create\>Dmflags or Create\>Advanced\>Dmflags?

NeonKnightOA:
Dmflags are numerous enough to guarantee their own page, separate from Advanced Options.

Alternatively, add the flags alongside the other options, and then the gametype-specific options in another column.

The-Gig:
I always meant to put dmflags in their own page (or maybe a two-columns layout for having Videoflags too?), also because more options may be added in the past (there is already the fast water movement one)... My question was about reaching it directly or passing through Advanced page first.
I meant whether you planned to place _the link_ to access dmflags page directly into "Create server" page or inside the (to be created, AFAIK) "Advanced" page? In the first way, it would require one click less to reach dmflags page, but would take up one more space in Create server page.

I'm not sure I found the right words to express it.

By the way, how should the link "case" look? DMFLAGS, dmflags, DMflags, DMFlags, Dmflags?

NeonKnightOA:
It should always be a direct entrance from Skirmish/Create Server. No one likes to navigate tons of pages solely to enable/disable an option.

NeonKnightOA:
Welp, I don't know what I did wrong, but it refuses to compile. I mean, the compiler doesn't complain (it may even pass a Github Action), but the linker _does_.

Anyway, [here's a .patch file](https://github.com/OpenArena/gamecode/files/14891286/advoptions.patch), download and apply with `git apply advoptions.patch`. And fix whatever's necessary. :/

NeonKnightOA:
I found out why I can't compile your code, and it's because ui_dmflags.c wasn't included in the Windows scripts.

However, it still doesn't compile. So here's [an updated patch](https://github.com/OpenArena/gamecode/files/14897372/advoptions-eddie.patch) adapted to the latest gamecode. Maybe you can figure out the error.

leilei-:
even if the gamecode got reverted in a fit, I am not opposed to having a dmflags menu (or any advanced subpages for server creation), so I am keeping this pull open

Issue #316, opened 2024-03-29
[Post-0.8.8 bug] Invisible Lightning from LG


NeonKnightOA:
One day, doing a run on the SP, I noticed that the Lightning Gun wasn't shooting anything but the ammo count still lowered and the marks on the walls and floors still kept being displayed.

I tested [the last 2023 nightly](https://files.poulsander.com/~poul19/public_files/oa/dev088/gamecode_nightly/openarena_gamecode_2023-05-30_905b539_2018-11-19_c719649_m.zip) and it still showed up.
I tested [the very first nightly](https://files.poulsander.com/~poul19/public_files/oa/dev088/gamecode_nightly/openarena_gamecode_2018-12-01_d967ac858bb7033ea06390b6a2977800f65a45ed.zip), it still showed up.
I tested with [b52](https://github.com/OpenArena/gamecode/releases/tag/oaxB52) and it still showed up.
I tested 0.8.8 and the LG behaved as normal.

So the bug may have been introduced between version 0.8.8 and one of the post-0.8.8 releases. Or perhaps it's a problem with my old config. I reset the config and the bug was gone. Still it's not a scenario that should be reproduced.

```
// generated by quake, do not modify
unbindall
bind TAB "+scores"
bind ENTER "+button2"
bind ESCAPE "togglemenu"
bind SPACE "+moveup"
bind + "sizeup"
bind - "sizedown"
bind 0 "weapon 10"
bind 1 "weapon 1"
bind 2 "weapon 2"
bind 3 "weapon 3"
bind 4 "weapon 4"
bind 5 "weapon 5"
bind 6 "weapon 6"
bind 7 "weapon 7"
bind 8 "weapon 8"
bind 9 "weapon 9"
bind = "sizeup"
bind _ "sizedown"
bind ` "toggleconsole"
bind a "weapon 8"
bind d "+back"
bind e "+forward"
bind f "+moveright"
bind g "weapon 5"
bind i "messagemode"
bind o "messagemode2"
bind q "weapon 10"
bind r "weapon 6"
bind s "+moveleft"
bind t "weapon 7"
bind w "weapon 11"
bind ~ "toggleconsole"
bind BACKSPACE "dropweapon"
bind PAUSE "pause"
bind CTRL "+movedown"
bind F1 "vote yes"
bind F2 "vote no"
bind F3 "ui_teamorders"
bind F11 "screenshot"
bind F12 "screenshotJPEG"
bind MOUSE1 "+attack"
[...]

PR #346, opened 2024-06-21
Fix secondary weapon models registering

NoiseByNorthwest:
Several people have noticed a curious regression which causes some secondary weapon models (such as machinegun's barrel) to not be rendered.
For instance:
- https://github.com/OpenArena/gamecode/issues/316
- https://github.com/OpenArena/engine/issues/85
- https://forum.manjaro.org/t/openarena-lightning-gun-beam-not-visible-with-glibc-2-37-x86-64-bug/135312

And some people have hypothesized that this could have something to do with a recent change in glibc.

I'm also facing this issue since an upgrade from ubuntu 22.04 to 24.04 and I've found its cause, here is the patch.

As you can see the logic to add a suffix to the primary weapon model was previously based on a call to COM_StripExtension() with the same pointer for in & out.
However, this function passes in & out to Q_strncpyz() which itself relies on strncpy() which specifies that (from C99 - ISO/IEC 9899:1999):

> If copying takes place between objects that overlap, the behavior is undefined.

So it was relying on an undefined behavior, prone to change at any time and without any communication.

This patch passes the build step but I've tested it in-game with the legacy project https://github.com/OpenArena/legacy and a back-ported version of it.

### DIFF
--- a/code/cgame/cg_weapons.c
+++ b/code/cgame/cg_weapons.c
@@ -855,8 +855,7 @@ void CG_RegisterWeapon( int weaponNum )
 		weaponInfo->ammoModel = trap_R_RegisterModel( ammo->world_model[0] );
 	}
 
-	Q_strncpyz( path, item->world_model[0], MAX_QPATH );
-	COM_StripExtension(path, path, sizeof(path));
+	COM_StripExtension(item->world_model[0], path, sizeof(path));
 	Q_strcat( path, sizeof(path), "_flash.md3" );
 	weaponInfo->flashModel = trap_R_RegisterModel( path );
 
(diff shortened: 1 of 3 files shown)


NoiseByNorthwest:
Regarding the overlapping pointers's behavior change of strncpy, according to https://forum.manjaro.org/t/openarena-lightning-gun-beam-not-visible-with-glibc-2-37-x86-64-bug/135312 it was introduced between 2.36 & 2.37, which is consistent with my observations: OK in 22.04 (glibc 2.35), KO in 24.04 (glibc 2.39).
I highly suspect these 2 commits https://github.com/bminor/glibc/commit/642933158e7cf072d873231b1a9bb03291f2b989 & https://github.com/bminor/glibc/commit/f049f52dfeed8129c11ab1641a815705d09ff7e8 as these patches are a major refactor of strncpy for x86's avx2/evex.

But, once again, it is not a glibc regression, but rather a bad use of strncpy (relying on an undefined behavior) by openarena code base.

The-Gig:
Do you know whether the problem also in q3a?
If the fix requires a gamecode update, I wonder how many mods are affected...

NoiseByNorthwest:
What do you mean by "q3a" ? I'm not yet familiar with internal details & organization of this project.
In case that answers your question, I've observed that this patch must be applied to the "gamecode" part (i.e. not the "engine" one) to be effective.

The-Gig:
I mean "quake 3 arena".
OpenArena is a free software clone of q3a, and OA gamecode is based upon the gpl release of q3a source code. OpenArena can run many mods which were initially created for q3... A fix in baseoa gamecode can't fix a bug in Bazooka Quake 3 mod, just to make an example.

I'm just wondering whether quake 3 and its mods are affected, too.

NoiseByNorthwest:
> I mean "quake 3 arena". OpenArena is a free software clone of q3a, and OA gamecode is based upon the gpl release of q3a source code. OpenArena can run many mods which were initially created for q3... A fix in baseoa gamecode can't fix a bug in Bazooka Quake 3 mod, just to make an example.

Sounds obvious once said

> I'm just wondering whether quake 3 and its mods are affected, too.

According to [ID Software's repository](https://github.com/id-Software/Quake-III-Arena/tree/master), no, the [COM_StripExtension() implentation is OK](https://github.com/id-Software/Quake-III-Arena/blob/master/code/game/q_shared.c#L61C24-L61C25) with aliased pointer arguments.

NoiseByNorthwest:
On OA side, the issue seems to be present from the beginning https://github.com/OpenArena/gamecode/commit/4a087ee80f2127ca6bd116038258cb5bbc1f5d47#diff-ed913d9e94c32881daa19f46560298ce10b1b520185ac53c40ca0911ae7df09eR86

With, from the beginning, a different implementation of COM_StripExtension().

The-Gig:
Sill wondering whether there are two different problems (one linux only already present in 0.8.8 and one also affecting windows somewhere in POST-0.8.8 gamecode -which however I wasn't able to reproduce, maybe that one might be graphics-card specific, idk- https://github.com/OpenArena/gamecode/issues/316#issuecomment-2092421046)...

And maybe pointing out the issue to gibc guys <https://sourceware.org/glibc/bugs.html> may still be useful, now that the affected function has been identified? Nobody knows when a new official OA gamecode package will be released (not in a short time I think), and oa-based mods would need to backport this fix too...

NoiseByNorthwest:
> Sill wondering whether there are two different problems (one linux only already present in 0.8.8 and one also affecting windows sonewhere in POST-0.8.8 gamecode -which however I wasn't able to reproduce, maybe that one might be graphics-card specific, idk- [#316 (comment)](https://github.com/OpenArena/gamecode/issues/316#issuecomment-2092421046))...

Windows + post 8.8.0 -> clearly another issue.

> And maybe pointing out the issue to gibc guys https://sourceware.org/glibc/bugs.html may still be useful, now that the affected function has been identified? Nobody knows when a new official OA gamecode package will be released (not in a short time I think), and oa-based mods would need to backport this fix too...


I don't think it would be useful, the glib guys wouldn't get why we are noticing them. What is important is that the issue is now clearly identified on OA side so that if it happens again on a mod it would be an easy fix.

PR #349, opened 2025-10-07
fix: shotgun not gibbing unless aiming at feet

WofWca:
Original issue: https://github.com/ioquake/ioq3/issues/794.
Original (rejected) MR: https://github.com/ioquake/ioq3/pull/795.
Same MR in baseq3a project:
https://github.com/ec-/baseq3a/pull/49.

Edit: for a video demo, see my comment below, https://github.com/OpenArena/gamecode/pull/349#issuecomment-3506110325

### DIFF
--- a/code/game/g_combat.c
+++ b/code/game/g_combat.c
@@ -834,12 +834,10 @@ void player_die( gentity_t *self, gentity_t *inflictor, gentity_t *attacker, int
 		// unless you aim at the feet.
 		// See https://github.com/ioquake/ioq3/issues/794.
 		//
-		// Note that without this line (when `g_shotgunGibFix` is enabled),
-		// when shooting at two players standing
-		// behind each other, the second target will take less damage,
-		// because the dead body of the first player will absorb the pellets
-		// until it gets gibbed (that is, up to 4 pellets,
-		// see `GIB_HEALTH` and `DEFAULT_SHOTGUN_DAMAGE`).
+		// Not executing this line makes is so that the corpse
+		// doesn't get shorter immediately on death
+		// and instead can still take up other pellets
+		// from the same shotgun shot.
 		//
 		// The purpose and the effect of this line is not entirely clear.
 		// Maybe it's to transition the player hitbox
--- a/code/game/g_combat.c
+++ b/code/game/g_combat.c
@@ -829,7 +829,26 @@ void player_die( gentity_t *self, gentity_t *inflictor, gentity_t *attacker, int
 
 	self->s.loopSound = 0;
 
-	self->r.maxs[2] = -8;
+	if (!g_shotgunGibFix.integer) {
+		// Executing this line causes a bug where the shotgun doesn't gib
+		// unless you aim at the feet.
+		// See https://github.com/ioquake/ioq3/issues/794.
+		//
+		// Note that without this line (when `g_shotgunGibFix` is enabled),
+		// when shooting at two players standing
+		// behind each other, the second target will take less damage,
+		// because the dead body of the first player will absorb the pellets
+		// until it gets gibbed (that is, up to 4 pellets,
+		// see `GIB_HEALTH` and `DEFAULT_SHOTGUN_DAMAGE`).
+		//
+		// The purpose and the effect of this line is not entirely clear.
+		// Maybe it's to transition the player hitbox
+		// into the "lying down dead" state, make it shorter.
+		// But this is already handled in `PM_CheckDuck`,
+		// so maybe it's just leftover code.
+		// So let's keep this line for now, but put behind a cvar.
+		self->r.maxs[2] = -8;
+	}
 
 	// don't allow respawn until the death anim is done
 	// g_forcerespawn may force spawning at some later time
--- a/code/game/g_local.h
+++ b/code/game/g_local.h
@@ -1068,6 +1068,7 @@ extern vmCvar_t g_votecustom;
 extern vmCvar_t g_warmup;
 extern vmCvar_t g_doWarmup;
 extern vmCvar_t g_blood;
+extern vmCvar_t g_shotgunGibFix;
 extern vmCvar_t g_allowVote;
 extern vmCvar_t g_teamAutoJoin;
 extern vmCvar_t g_teamForceBalance;
--- a/code/game/g_main.c
+++ b/code/game/g_main.c
@@ -79,6 +79,7 @@ vmCvar_t g_restarted;
 vmCvar_t g_logfile;
 vmCvar_t g_logfileSync;
 vmCvar_t g_blood;
+vmCvar_t g_shotgunGibFix;
 vmCvar_t g_podiumDist;
 vmCvar_t g_podiumDrop;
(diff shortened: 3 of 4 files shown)


The-Gig:
I checked, and it's actually true. You can gib a player by shooting him with the shotgun in the feet, but not in the body. What an odd small bug.
I see there you also mentioned something about knockback. Is that different between shooting at body and feet, too?

WofWca:
> I see there you also mentioned something about knockback. Is that different between shooting at body and feet, too?

My point about knockback was that you only get knockback for one pellet, i.e. 10 damage, and not all the damage, i.e. shotgun's knockback is much weaker than that of the railgun, for example.
As long as the player is alive, the knockback is always the same, regardless of where you aim. But maybe (I'm not sure), if you aim at the feet, the dead body will be knocked back further. Although this is not significant, because it's just 10 damage worth of knockback.

The-Gig:
About knockback, I just did a test with OA 0.8.8 and it looks to me that how much a shotgun hit "pushes" a character _already_ changes depending from how accurate your shot is.

My steps:
- Launched the game in windowed mode and loaded a map with /devmap command (cheats enabled).
- Started a second openarena client on a different window, and used Multiplayer menu to join the match hosted by the other instance.
- Used /god cheat to make one of the characters invulnerable to weapon damage.
- Fired the shotgun at the "god" character, from both point blank and long range.
It looked to me that the point blank shot pushed him further back than the long range shot. It doesn't seem to scale linearly to me, but *I guess* that's on purpose otherwise it would be too much.
- Set /g_knockback 10000 (default is 1000) to make it clearer.
- Fired the shotgun at the "god" character again, from both point blank and long range.
It seemed to me even more evident that point blank shots push the character further back than long range shots.

Note: I just aimed at the opponent, without expressly aiming at the legs. However it's possible some pellets hit the lower part of the model. I don't know whether upper or lower part of the hitbox makes any difference.

WofWca:
Hmm, you are right.
Sorry, pehaps I mixed up the actual velocity knockback and the timer. The velocity seems to be getting added up here

https://github.com/OpenArena/gamecode/blob/d371ff1482d9c00c543b29186d26d4d875b3bb0a/code/game/g_combat.c#L1168

But the timer here only gets set by one pellet:

https://github.com/OpenArena/gamecode/blob/d371ff1482d9c00c543b29186d26d4d875b3bb0a/code/game/g_combat.c#L1173-L1183

In that case, this MR still only changes how dead players behave.

The-Gig:
I just noticed in this commit https://github.com/ec-/baseq3a/pull/49/files/ad81db1b27e669c7c59aca8b1c22e6c552d1a2ac..0efcf57b4d3c5b6b794cced74864b079e15e7abe
the notes about the fact that the behaviour changes like providing fewer damage to someone behind a dead body...
Could you please explain that better to me? Thanks...

WofWca:
Yes sure. Here is a visual example. First, without the proposed fix, i.e. how things are currently:

https://github.com/user-attachments/assets/20a1efe9-1548-4cc3-a921-eacc9e6d8d68

image

And now _with_ the proposed fix:

https://github.com/user-attachments/assets/23bf3e14-1343-4271-931e-ab8860e94eb7

image

Notice that now the first Crash (her dead body, rather) eats up 4 more pellets, and the second Crash gets only 2 out of 11, and thus survives. It took 5 pellets to kill the first crash, but she took 9 pellets.

The-Gig:
I see. Then, changing it may be considered changing part of the Q3 gameplay, despite being quite a side case. It _might_ be an intentional choice by idsoftware instead of a bug, with a reasoning like "the shotgun is already underpowered comparing to other weapons, no need to waste pellets on dead bodies unless someone really wants to shoot the body on the floor". So, to be sure to do not fall under OA "NOTTODO", I guess one may make the fix disabled by default and controlled by an apposite cvar (or a DMFLAGS value, to avoid wasting precious cvar memory on details) as you did... or maybe having a different approach and make those pellets both hit the corpse AND one potential player behind it (like passing through the dead or gibbed body from a side to the other)?
I'm still curious to know what happens with the nailgun (or other weapons) in similar situations...

WofWca:
> "the shotgun is already underpowered comparing to other weapons, no need to waste pellets on dead bodies unless someone really wants to shoot the body on the floor"

Considering that they nerfed it in Quake Live (I think it deals 80 damage now), I doubt that this is the case.

> or maybe having a different approach and make those pellets both hit the corpse AND one potential player behind it (like passing through the dead or gibbed body from a side to the other)?

I think that the bug that this MR fixes cannot be fixed without affecting gameplay at all. A dead body is not purely decorative, it can shield from bullets / pellets, e.g. when it's upstairs and you try to shoot someone through it.

>  I'm still curious to know what happens with the nailgun (or other weapons) in similar situations...

Such weapons can't perform multiple shots in one frame, so I'm pretty sure that they're basically unaffected. Well, maybe except for the case where multiple players make a shot in the same frame?

> or a DMFLAGS value

That's something that's sent from the server to the client, right? The MR only affects the server, there is no need to change the network protocol I think.

> I guess one may make the fix disabled by default and controlled by an apposite cvar

OK, I am fine with merging this disabled by default for a start, enabling can be discussed separately.

WofWca:
> about the fact that the behaviour changes like providing fewer damage to someone behind a dead body

Continuing: with `com_blood 0` with this fix the first player will actually eat up _all_ the pellets, not just 4.

The-Gig:
> > about the fact that the behaviour changes like providing fewer damage to someone behind a dead body
> 
> Continuing: with `com_blood 0` with this fix the first player will actually eat up _all_ the pellets, not just 4.

That sounds like a problem. I'm starting to wonder whether this fix is worth the effort.
Maybe it's even more correct to call it an "alternate rule" instead of "fix", as it doesn't seem to come from a bug, but from a deliberate design choice.
I admit gibbing on fragging is satisfactory, but if it comes at the cost of your weapon being less efficient in a crowd fray? IDK.
I may still be ok with it considering it's not enabled by default.

WofWca:
> as it doesn't seem to come from a bug, but from a deliberate design choice.

I still doubt that. This would have been easy to fix otherwise, e.g. inside of `ShotgunPellet`. There is already code to make pellets pass through multiple entities (see the `for` loop) so it's easy to keep the pellet if it's passing through a dead body:

https://github.com/OpenArena/gamecode/blob/025d2838a609781d18655e0a55a73823f73ae8ca/code/game/g_weapon.c#L344-L389

In fact, I can implement that. In that case the shotgun will technically get a little _more_ powerful, for the rare case of shooting someone through someone else's feet.

WofWca:
OK, I pushed another commit. As you can see, now pellets go through dead bodies, as before. On the second screenshot, you can see that the first Crash absorbed 5 pellets before they started hitting the second Crash.

https://github.com/user-attachments/assets/88854c7d-9abd-4661-8e74-20c219206792

image

https://github.com/user-attachments/assets/77b91d05-f61d-4059-90cb-0df9fe4ca6da

image

leilei-:
superfluous

